PR #7, opened 2015-10-08
Hive import

Ishiihara:
This PR adds necessary utilities to import files in HDFS to Hive. 
@ewencp Can you take a look at it when you hive time? It's based on the WAL branch, but all the changes related to Hive are in the hive directory.

### DIFF
--- a/src/main/java/io/confluent/copycat/hdfs/hive/HiveExec.java
+++ b/src/main/java/io/confluent/copycat/hdfs/hive/HiveExec.java
@@ -1,0 +1,139 @@
+/**
+ * Copyright 2015 Confluent Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+package io.confluent.copycat.hdfs.hive;
+
+import org.apache.avro.Schema;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.cli.CliDriver;
+import org.apache.hadoop.hive.cli.CliSessionState;
+import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.ql.session.SessionState;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+
+import io.confluent.copycat.hdfs.HdfsSinkConnectorConfig;
+
+/**
+ * Utility to create an external table for an directory with provided schema.
+ */
+public class HiveExec {
+
+  public static final Log log = LogFactory.getLog(HiveExec.class);
+
+  private Configuration conf;
+  private HiveConf hiveConf;
+  private CliDriver cliDriver;
+  private final HiveQLWriter hiveQLWriter;
+  private static final String HIVE_SASL_ENABLED = "hive.metastore.sasl.enabled";
+
+  /**
+   * HiveExec constructor
+   * @param conf Hadoop configuration
+   * @param config HDFS Connector configuration
+   */
+  public HiveExec(Configuration conf, HdfsSinkConnectorConfig config) {
+    this.conf = conf;
+    String topicsDir = config.getString(HdfsSinkConnectorConfig.TOPIC_DIR_CONFIG);
+    String url = config.getString(HdfsSinkConnectorConfig.HDFS_URL_CONFIG);
+    hiveConf = new HiveConf();
+    String hiveConfDir = config.getString(HdfsSinkConnectorConfig.HIVE_CONF_DIR_CONFIG);
+    hiveConf.addResource(new Path(hiveConfDir, "hive-site.xml"));
+    SessionState.start(new CliSessionState(hiveConf));
+    cliDriver = new CliDriver();
+    hiveQLWriter = new HiveQLWriter(url, topicsDir);
+  }
+
+  /**
+   * Create an Avro external table
+   * @param tableName the name of the table to create in Hive.
+   * @param avroSchema the schema of the Avro files in HDFS.
+   */
+  public void createAvroTable(String tableName, Schema avroSchema) throws IOException {
+    // generate the HiveQL statements to run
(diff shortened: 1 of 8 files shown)


src/main/java/io/confluent/copycat/hdfs/hive/HiveExec.java
```diff
@@ -0,0 +1,135 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+import io.confluent.copycat.hdfs.HdfsSinkConnectorConfig;
+
+/**
+ * Utility to create an external table for an directory in .
```
ewencp: Incomplete sentence

src/main/java/io/confluent/copycat/hdfs/hive/HiveExec.java
```diff
@@ -0,0 +1,135 @@
...
+  private final HiveQLWriter hiveQLWriter;
+  private static final String HIVE_SASL_ENABLED = "hive.metastore.sasl.enabled";
+
+  public HiveExec(Configuration conf, HdfsSinkConnectorConfig config) {
+    this.conf = conf;
+    String topicsDir = config.getString(HdfsSinkConnectorConfig.TOPIC_DIR_CONFIG);
+    String url = config.getString(HdfsSinkConnectorConfig.HDFS_URL_CONFIG);
+    hiveConf = new HiveConf();
+    String hiveConfDir = config.getString(HdfsSinkConnectorConfig.HIVE_CONF_DIR_CONFIG);
+    hiveConf.addResource(new Path(hiveConfDir, "hive-site.xml"));
```
ewencp: Is this filename hardcoded in Hive or just a default?

src/main/java/io/confluent/copycat/hdfs/hive/HiveExec.java
```diff
@@ -0,0 +1,135 @@
...
+    String createAvroTableStmt = hiveQLWriter.createAvroTableStmt(tableName, avroSchema) + ";\n";
+    executeQuery(createAvroTableStmt);
+    log.info("Create Avro table complete.");
+  }
+
+  /**
+   * @param tableName the name of the table to create in Hive.
+   * @param avroSchema the schema of the Avro files in HDFS.
+   */
+  public void alterScheam(String tableName, Schema avroSchema) throws IOException {
```
ewencp: Typo: alterScheam -> alterSchema

src/main/java/io/confluent/copycat/hdfs/hive/HiveExec.java
```diff
@@ -0,0 +1,135 @@
...
+      }
+    } catch (Exception e) {
+      throw new IOException("Exception thrown in Hive", e);
+    }
+  }
+
+  public String[] getHiveArgs(String... args) throws IOException {
+    List<String> newArgs = new LinkedList<>();
+    newArgs.addAll(Arrays.asList(args));
+    addHiveConfigs(hiveConf, conf);
```
ewencp: Why not just add these hive configs once? It seems like hiveConf and conf are always going to be the same since they aren't parameters to this method.

src/test/java/io/confluent/copycat/hdfs/hive/HiveExecTest.java
```diff
@@ -0,0 +1,164 @@
...
+    Schema avroSchema = AvroData.fromCopycatSchema(createSchema());
+    HiveExec hiveExec = new HiveExec(conf, connectorConfig);
+
+    runHive(hiveExec, "DROP TABLE " + topic);
+
+    hiveExec.createAvroTable(topic, avroSchema);
+    hiveExec.addPartition(topic, 0);
+    hiveExec.setPartitionLoc(topic, 0);
+
+    String[] expectedResult = {"true", "12", "12", "12.2", "12.2", "0"};
```
ewencp: What is the `"0"` at the end from? Aren't there only 5 columns?

src/test/java/io/confluent/copycat/hdfs/hive/HiveExecTest.java
```diff
@@ -0,0 +1,164 @@
...
+    String result = runHive(hiveExec, "SELECT * from " + topic);
+    String[] rows = result.split("\n");
+    assertEquals(7, rows.length);
+    for (String row: rows) {
+      String[] parts  = row.replace(" ", "").split("\t");
```
ewencp: If we keep doing this, we might just want to refactor it into a method. It's not confusing or anything, but just for clarity.

Also, will this work if any of the returned items are strings?

src/test/java/io/confluent/copycat/hdfs/hive/HiveExecTest.java
```diff
@@ -0,0 +1,164 @@
...
+  @Test
+  public void testAlterSchema() throws Exception {
+    String topic = "test";
+    Schema avroSchema = AvroData.fromCopycatSchema(createSchema());
+    HiveExec hiveExec = new HiveExec(conf, connectorConfig);
+
+    runHive(hiveExec, "DROP TABLE " + topic);
+    hiveExec.createAvroTable(topic, avroSchema);
+
+    Schema newAvroSchema = AvroData.fromCopycatSchema(createNewSchema());
```
ewencp: Is it worth testing this with any data? Both before and after altering the table?

src/test/java/io/confluent/copycat/hdfs/hive/HiveExecTest.java
```diff
@@ -0,0 +1,164 @@
...
+    String[] lines = description.split("\n");
+
+    for (int i = 0; i < colTypes.length; ++i) {
+      String[] parts = lines[i].replace(" ", "").split("\t");
+      assertEquals(colTypes[i], parts[1]);
+    }
+  }
+
+  private String runHive(HiveExec hiveExec, String query) throws Exception {
+    ByteArrayOutputStream outBytes = new ByteArrayOutputStream();
```
ewencp: I took a quick look at the CliDriver and it looks like it just hard codes things so output goes to stdout/stderr. There's not an alternative that gives you any better control than this? It's annoying, but not horrible, in tests. But doesn't the fact that we're doing this here mean that we're also going to get unformatted output to stdout during execution of Hive queries during normal operation? It seems like that's going to make a mess of logs and also means that potentially important output is lost to stdout instead of making it through the log4j system.

src/main/java/io/confluent/copycat/hdfs/hive/HiveExec.java
```diff
@@ -0,0 +1,140 @@
...
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+
+import io.confluent.copycat.hdfs.HdfsSinkConnectorConfig;
+
+/**
+ * Utility to create an external table for an directory with a desired hive .
```
ewencp: Very minor, but this comment is incomplete.

ewencp:
@Ishiihara This LGTM for what it does. I assume it was intentional that we haven't made any calls from the tasks to HiveExec yet? Presumably since some of the calls need the schema projection patch as well (e.g. the ones that alter table schemas)?

PR #8, opened 2015-10-26
Schema

Ishiihara:
Add schema projection for different schema compatibility.

### DIFF
--- a/src/main/java/io/confluent/copycat/hdfs/AvroRecordWriterProvider.java
+++ b/src/main/java/io/confluent/copycat/hdfs/AvroRecordWriterProvider.java
@@ -41,7 +41,8 @@ public class AvroRecordWriterProvider implements RecordWriterProvider {
 
     final Schema schema = record.valueSchema();
     final FSDataOutputStream out = path.getFileSystem(conf).create(path);
-    writer.create(avroData.fromCopycatSchema(schema), out);
+    org.apache.avro.Schema avroSchema = avroData.fromCopycatSchema(schema);
+    writer.create(avroSchema, out);
 
     return new RecordWriter<Long, SinkRecord>(){
       @Override
--- a/src/main/java/io/confluent/copycat/hdfs/HdfsWriter.java
+++ b/src/main/java/io/confluent/copycat/hdfs/HdfsWriter.java
@@ -14,12 +14,21 @@
 
 package io.confluent.copycat.hdfs;
 
+import org.apache.avro.file.DataFileReader;
+import org.apache.avro.file.FileReader;
+import org.apache.avro.file.SeekableInput;
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.io.DatumReader;
+import org.apache.avro.mapred.FsInput;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.PathFilter;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.copycat.data.Schema;
+import org.apache.kafka.copycat.data.SchemaProjector;
 import org.apache.kafka.copycat.errors.CopycatException;
 import org.apache.kafka.copycat.errors.IllegalWorkerStateException;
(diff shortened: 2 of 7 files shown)


Ishiihara:
@ewencp It is probably easier to review after #6 is merged. However, the main changes are in `HdfsWriter` and `HdfsWriterTest`.

src/main/java/io/confluent/copycat/hdfs/AvroRecordWriterProvider.java
```diff
@@ -41,7 +41,10 @@
 
     final Schema schema = record.valueSchema();
     final FSDataOutputStream out = path.getFileSystem(conf).create(path);
-    writer.create(avroData.fromCopycatSchema(schema), out);
+    org.apache.avro.Schema avroSchema = avroData.fromCopycatSchema(schema);
+    int version = schema.version();
```
ewencp: Be careful here, version isn't actually required so this could result in a null pointer exception.

src/main/java/io/confluent/copycat/hdfs/AvroRecordWriterProvider.java
```diff
@@ -41,7 +41,10 @@
...
+    avroSchema.addProp("version", String.valueOf(version));
```
ewencp: The `AvroData` class should already be saving the version information to a property called `copycat.version`. You could probably just reuse that instead.

src/main/java/io/confluent/copycat/hdfs/HdfsWriter.java
```diff
@@ -61,6 +70,9 @@
   private long backOffMs;
   private Set<TopicPartition> lastAssignment;
   private Map<TopicPartition, Long> failureTime;
+  private Map<TopicPartition, Schema> schemas;
+  private Map<TopicPartition, Integer> versions;
```
ewencp: Heh, the number of maps is really getting out of control here. We should definitely follow up sometime soon with a cleanup of these.

src/main/java/io/confluent/copycat/hdfs/HdfsWriter.java
```diff
@@ -519,4 +589,28 @@ public RecordWriter getRecordWriter(TopicPartition topicPart) {
...
+    FileStatus[] statuses = storage.listStatus(path, filter);
+    assert statuses.length == 1;
+
+    SeekableInput input = new FsInput(statuses[0].getPath(), conf);
+    DatumReader<Object> reader = new GenericDatumReader<>();
+    FileReader<Object> fileReader = DataFileReader.openReader(input, reader);
+    org.apache.avro.Schema avroSchema = fileReader.getSchema();
+    int version = Integer.parseInt(avroSchema.getProp("version"));
+    Schema schema = avroData.toCopycatSchema(avroSchema);
+    versions.put(topicPart, version);
```
ewencp: Isn't the version in the schema anyway? I found the two maps a bit confusing at first, and now I think they may be redundant?

src/main/java/io/confluent/copycat/hdfs/HdfsWriter.java
```diff
@@ -224,6 +260,8 @@ private void execute(TopicPartition topicPart) throws IOException {
       } catch (IllegalWorkerStateException e) {
         // Should we retry in this case?
         throw e;
+      } catch (SchemaProjectorException e) {
+        throw e;
```
ewencp: Why this separate block for `SchemaProjectorException`? Isn't that just a CopycatException that could be handled by the existing block? Or are you throwing the exception because this is effectively unrecoverable?

ewencp:
@Ishiihara Looks good. Left some comments for things that could be cleaned up/simplified. I think the only things that are critical to be addressed are:

1. Schemas that don't have versions, and therefore could a) be null when you ask for the schema and b) you need to actually check schema equality rather than just relying on the version numbers.
2. We need to better understand the behavior that should happen when we hit a SchemaProjectorException.

Ishiihara:
@ewencp Addressed the review comments. Can you take a look when you have time?

ewencp:
LGTM

PR #9, opened 2015-11-02
Custom partitioner

Ishiihara:
This PR refactors `HdfsWriter` to a collection of `TopicPartitionWriters` and adds `FieldPartitioner` and `TimeBasedPartitioner`.


src/main/java/io/confluent/copycat/hdfs/CommittedFileFilter.java
```diff
@@ -17,8 +17,12 @@
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathFilter;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 
 public class CommittedFileFilter implements PathFilter {
+  private static Pattern pattern = Pattern.compile("\\w+_\\d+_\\d+_\\d+");
```
ewencp: I can never remember the exact set of legal chars for topics, but I don't think `\w` will cover it. For example `-` should be a valid char I think. If you check in Kafka's code, I think there's a unit test that checks valid names, that's probably a good source to get the definitive list of valid characters.

src/main/java/io/confluent/copycat/hdfs/DataWriter.java
```diff
@@ -0,0 +1,223 @@
...
+      TopicPartition tp = new TopicPartition(topic, partition);
+      bucketWriters.get(tp).buffer(record);
+    }
+
+    for (TopicPartition tp: assignment) {
+      bucketWriters.get(tp).write();
+    }
+  }
+
+  public void recover() {
```
ewencp: This is never used, only the `recover(TopicPartition)` version is.

src/main/java/io/confluent/copycat/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -25,41 +25,86 @@
...
   private static final String RETRY_BACKOFF_DOC = "The retry backoff in milliseconds.";
   public static final long RETRY_BACKOFF_DEFALUT = 5000L;
+
   public static final String HIVE_HOME_CONFIG = "hive.home";
   private static final String HIVE_HOME_DOC = "Hive home directory";
   public static final String HIVE_CONF_DIR_CONFIG = "hive.conf.dir";
   private static final String HIVE_CONF_DIR_DOC = "Hive configuration directory";
 
+  public static final String PARTITIONER_CLASS_CONFIG = "partitioner.class";
+  private static final String PARTITIONER_CLASS_DOC = "The partitioner class.";
```
ewencp: This probably warrants a bit more explanation...

src/main/java/io/confluent/copycat/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -25,41 +25,86 @@
...
+  public static final String PARTITIONER_CLASS_DEFALUT = "io.confluent.copycat.hdfs.partitioner.DefaultPartitioner";
+
+  public static final String PARTITION_FIELD_CONFIG = "partition.field";
```
ewencp: Not a problem for this round, but something to think about refactoring in the future -- you can separate out configs for different classes into their own AbstractConfig subclasses. That's one way to get around a) having all the configs in one place and b) having to manually validate fields that are required for only some of the partitioners. Of course, the drawback to not having all the configs in one class is that automatically generating docs will be a bit harder.

src/main/java/io/confluent/copycat/hdfs/TopicPartitionWriter.java
```diff
@@ -0,0 +1,461 @@
...
+  private Set<String> appended;
+  private long offset = -1;
+  private Map<String, Long> startOffsets;
+  private Map<String, Long> offsets;
+  private long timeoutMs;
+  private long failureTime;
+
+  private Pattern pattern = Pattern.compile("\\w+_\\d+_\\d+_\\d+");
+
+  public TopicPartitionWriter(String url,
```
ewencp: Nice, I like that the giant class is getting refactored to break it down into smaller pieces that also clarify how data is organized & processed.

src/main/java/io/confluent/copycat/hdfs/TopicPartitionWriter.java
```diff
@@ -0,0 +1,461 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import io.confluent.copycat.avro.AvroData;
+import io.confluent.copycat.hdfs.partitioner.Partitioner;
+
+public class TopicPartitionWriter {
+  private static final Logger log = LoggerFactory.getLogger(TopicPartitionWriter.class);
+  private WAL wal;
+  private Map<String, String> tempfiles;
```
ewencp: I'm a bit confused how this works with more than one log file. In the case of time based partitioning, should there ever be more than one tempfile at a time? From what I can tell in this file, the partitioner isn't involved in deciding when to rotate files, so I'm not sure how we can correctly ensure exactly once delivery. Won't multiple open files here potentially break the way the WAL works?

Also, if the Partitioner isn't involved in choosing when to rotate, couldn't we end up with pretty bad file sizes? E.g. if the partitioner splits the data across many files, won't the files end up being quite small.

src/main/java/io/confluent/copycat/hdfs/TopicPartitionWriter.java
```diff
@@ -0,0 +1,461 @@
...
+  private Configuration conf;
+  private AvroData avroData;
+  private Set<String> appended;
+  private long offset = -1;
+  private Map<String, Long> startOffsets;
+  private Map<String, Long> offsets;
+  private long timeoutMs;
+  private long failureTime;
+
+  private Pattern pattern = Pattern.compile("\\w+_\\d+_\\d+_\\d+");
```
ewencp: I saw this pattern before, any place we could store it just once? Especially since I think it needs tweaking, so they are at risk of getting out of sync with each other.

src/main/java/io/confluent/copycat/hdfs/TopicPartitionWriter.java
```diff
@@ -0,0 +1,461 @@
...
+    }
+  }
+
+  public void write() {
+    long now = System.currentTimeMillis();
+    if (failureTime > 0 && now - failureTime < timeoutMs) {
+      return;
+    }
+    if (state.compareTo(State.WRITE_STARTED) < 0) {
+      recover();
```
ewencp: Just noticed that I think we missed something in the previous patch. When reusing `recover()` here, we need it to return a boolean or something -- if it does fail, then we should not continue to the loop below this, right? I think we should just be returning in that case.

src/main/java/io/confluent/copycat/hdfs/TopicPartitionWriter.java
```diff
@@ -0,0 +1,461 @@
...
+    if (writers.containsKey(encodedPartition)) {
+      RecordWriter writer = writers.get(encodedPartition);
+      writer.close();
+      writers.remove(encodedPartition);
+    }
+    recordCounter = 0;
+  }
+
+  private void closeTempfile() throws IOException {
+    for (String encodedPartition: tempfiles.keySet()) {
```
ewencp: I think I see the answer to my earlier question. Isn't this approach a problem if we write ? I think this patch basically makes it so we need atomic commits of multiple files since data can be spread across multiple files now. So we either need each entry in the WAL to include the list of renames to perform *or* we need some sort of start/end transaction records that indicate when all the information is available. Otherwise, `appendToWAL` might only manage to append a half the files and you'll deliver only half the data you expected to. As long as it's all guaranteed to be recorded in the WAL *or* we won't apply the half-completed write, it should be fine, but we need that guarantee.

By the way, this might be a good way to write a system test that stresses the WAL -- use a partitioning that causes a *lot* of files to be used in the same transaction. That'll drastically increase the likelihood of breaking the WAL/deliver guarantees.

src/main/java/io/confluent/copycat/hdfs/TopicPartitionWriter.java
```diff
@@ -0,0 +1,461 @@
...
+    long startOffset = startOffsets.get(encodedPartition) + 1;
+    long endOffset = offsets.get(encodedPartition) + 1;
+    String directory = getDirectory(encodedPartition);
+    String committedFile = FileUtils.committedFileName(url, topicsDir, directory, tp, startOffset, endOffset);
+    wal.append(tempfile, committedFile);
+    appended.add(tempfile);
+  }
+
+  private void appendToWAL() throws IOException {
+    for (String encodedPartition: tempfiles.keySet()) {
```
ewencp: I think I see the answer to my earlier question. Isn't this approach a problem if we write ? I think this patch basically makes it so we need atomic commits of multiple files since data can be spread across multiple files now. So we either need each entry in the WAL to include the list of renames to perform *or* we need some sort of start/end transaction records that indicate when all the information is available. Otherwise, `appendToWAL` might only manage to append a half the files and you'll deliver only half the data you expected to. As long as it's all guaranteed to be recorded in the WAL *or* we won't apply the half-completed write, it should be fine, but we need that guarantee.

By the way, this might be a good way to write a system test that stresses the WAL -- use a partitioning that causes a *lot* of files to be used in the same transaction. That'll drastically increase the likelihood of breaking the WAL/deliver guarantees.

src/main/java/io/confluent/copycat/hdfs/TopicPartitionWriter.java
```diff
@@ -0,0 +1,461 @@
...
+    }
+    long startOffset = startOffsets.get(encodedPartition) + 1;
+    long endOffset = offsets.get(encodedPartition) + 1;
+    String directory = getDirectory(encodedPartition);
+    String committedFile = FileUtils.committedFileName(url, topicsDir, directory, tp, startOffset, endOffset);
+    wal.append(tempfile, committedFile);
+    appended.add(tempfile);
+  }
+
+  private void appendToWAL() throws IOException {
```
ewencp: See above comment -- this method definitely needs some sort of adjustment to ensure exactly once delivery.

src/main/java/io/confluent/copycat/hdfs/partitioner/DailyPartitioner.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import io.confluent.common.config.ConfigException;
+import io.confluent.copycat.hdfs.HdfsSinkConnectorConfig;
+
+public class DailyPartitioner extends TimeBasedPartitioner {
+
+  private static long partitionDurationMs = TimeUnit.HOURS.toMillis(24);
+  private static String pathFormat = "'daily'/YYYY/MM/dd/";
```
ewencp: I'm confused about what this string is doing. Is the `'daily'` supposed to be in there like that? And have the tick marks around it? Is that just joda time formatting? And if that's prefixing with a `daily` directory, is that what users would actually want or do they only want the `YYYY/MM/dd` part?

src/main/java/io/confluent/copycat/hdfs/partitioner/FieldPartitioner.java
```diff
@@ -0,0 +1,88 @@
...
+    Schema valueSchema = sinkRecord.valueSchema();
+    if (value instanceof Struct) {
+      Struct struct = (Struct) value;
+      Object partitionKey = struct.get(fieldName);
+      Type type = valueSchema.field(fieldName).schema().type();
+      switch (type) {
+        case INT8:
+        case INT16:
+        case INT32:
+        case INT64:
```
ewencp: float32, float64? Not sure how complete these needs to be. Long term, it might make sense to allow for settings like `PARTITION_FIELD_CONFIG = "user.name"` which is automatically turned into field/map entry lookups. Certainly ok to skip for now, but might be worth looking into what other systems support wrt partitioning to see whether something like that would be useful.

src/main/java/io/confluent/copycat/hdfs/partitioner/Partitioner.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.Map;
+
+/**
+ * Partition incoming records, and generates directories and file names in which to store the
+ * incoming records.
+ */
+public interface Partitioner {
+  void configure(Map<String, Object> config);
+  String encodePartition(SinkRecord sinkRecord);
```
ewencp: Can we add some documentation here? This is actually flexible enough and a nice extension point that people may want to implement their own Partitioners to plug in, especially as they would otherwise have to wait for updated releases. A small amount of guidance could go a long way here.

src/main/java/io/confluent/copycat/hdfs/partitioner/Partitioner.java
```diff
@@ -0,0 +1,30 @@
...
+  String encodePartition(Object partitionKey);
```
ewencp: This doesn't seem to be used anywhere except in a couple of tests. Can we get rid of it and the implementations?

src/main/java/io/confluent/copycat/hdfs/partitioner/Partitioner.java
```diff
@@ -0,0 +1,30 @@
...
+  String generatePartitionedPath(String topic, String encodedPartition);
```
ewencp: Can you explain why we need both `encodePartition(SinkRecord)` and `generatePartitionedPath`? It seems like if we generated the directory name we wanted to place the data under, that should be enough. Both are strings anyway so they seem equivalent.

More generally, the issue I mentioned about daylight savings time -- I'm wondering if it would be more reliable to have joda-time handle extracting just the date info and formatting it without us having to handle dividing the timeline up into even pieces. In particular, I'm worried that for something like the hourly partitioner, we currently might be generating the same directory name for two different blocks of time when daylight savings time hits (when the clock rolls back, we see the hour 2am - 3am 2x, but the unix timestamps are different). I'm honestly now sure how people would prefer for this to be handled....

src/main/java/io/confluent/copycat/hdfs/partitioner/TimeBasedPartitioner.java
```diff
@@ -0,0 +1,103 @@
...
+  protected void init(long partitionDurationMs, String pathFormat, Locale locale, DateTimeZone timeZone) {
+    this.partitionDurationMs = partitionDurationMs;
+    this.formatter = getDateTimeFormatter(pathFormat, timeZone).withLocale(locale);
+  }
+
+  private static DateTimeFormatter getDateTimeFormatter(String str, DateTimeZone timeZone) {
+    return DateTimeFormat.forPattern(str).withZone(timeZone);
+  }
+
+  private static long getPartition(long timeGranularityMs, long timestamp) {
```
ewencp: This method isn't used anywhere.

src/test/java/io/confluent/copycat/hdfs/partitioner/HourlyPartitionerTest.java
```diff
@@ -0,0 +1,46 @@
...
+public class HourlyPartitionerTest {
+
+  private Map<String, Object> createConfig() {
+    Map<String, Object> config = new HashMap<>();
+    config.put(HdfsSinkConnectorConfig.LOCALE_CONFIG, "en");
+    config.put(HdfsSinkConnectorConfig.TIMEZONE_CONFIG, "America/Los_Angeles");
+    return config;
+  }
+
+  @Test
```
ewencp: This is where we might a test for daylight savings time.

ewencp:
@Ishiihara This looks like it's very much in the right direction, the major issue I have is that I think the WAL implementation doesn't work correctly with multiple files.

@gwenshap Would be helpful to have your eyes on this as well. Not necessarily the whole patch, but that we're hitting the type of partitioning that users will find useful (and maybe a review to make sure any caveats around their functionality, like the DST issues I mentioned in the review for the DailyPartitioner).

Ishiihara:
@ewencp Sorry this turned out to be a much larger patch than I originally planned. Added the Hive execution service and some parquet related tests.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -0,0 +1,333 @@
...
+  private Map<TopicPartition, Long> offsets;
+  private Set<TopicPartition> lastAssignment;
+  private HdfsSinkConnectorConfig connectorConfig;
+  private AvroData avroData;
+  private SinkTaskContext context;
+  private ExecutorService executorService;
+  private String hiveDatabase;
+  private HiveMetaStore hiveMetaStore;
+  private HiveUtil hive;
+  private Queue<Future> futures;
```
ewencp: Futures for what? This names could probably be clearer -- maybe something like `hiveUpdateFutures`?

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -0,0 +1,333 @@
...
+      writerProvider = format.getRecordWriterProvider();
+      schemaFileReader = format.getSchemaFileReader(avroData);
+
+      partitioner = createPartitioner(connectorConfig);
+
+      assignment = context.assignment();
+      offsets = new HashMap<>();
+      lastAssignment = new HashSet<>();
+
+      hiveIntegration = connectorConfig.getBoolean(HdfsSinkConnectorConfig.HIVE_INTEGRATION_CONFIG);
```
ewencp: Why do we need a separate boolean flag here to indicate whether Hive integration should be turned on? Shouldn't the existence of a Hive DB config be enough to tell whether you want Hive support?

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -0,0 +1,333 @@
...
+  public void write(Collection<SinkRecord> records) {
+    for (SinkRecord record: records) {
+      String topic = record.topic();
+      int partition = record.kafkaPartition();
+      TopicPartition tp = new TopicPartition(topic, partition);
+      topicPartitionWriters.get(tp).buffer(record);
+    }
+
+    if (hiveIntegration) {
+      Iterator<Future> iterator = futures.iterator();
```
ewencp: Would it make sense to do this in a callback tied to the future (e.g. see `FutureCallback` in Kafka Connect's utility library, that implements both a callback and a future)? It's not that big a deal to make it work as is, but it's basically polling on every `write` call rather than just being event-driven by removing the future immediately upon completion.

That said, maybe the current approach is just simpler since it avoids any additional locking? I assume the future is completed in the `executorService` thread.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -25,47 +25,115 @@
 
   public static final String HDFS_URL_CONFIG = "hdfs.url";
   private static final String HDFS_URL_DOC = "HDFS connection URL.";
-  public static final String RECORD_WRITER_PROVIDER_CLASS_CONFIG = "record.writer.provider.class";
-  private static final String RECORD_WRITER_PROVIDER_CLASS_DOC = "";
+
+  public static final String FORMAT_CONFIG = "format";
+  private static final String FORMAT_DOC = "The format class to use when writing data to HDFS.";
+  public static final String FORMAT_DEFAULT = "io.confluent.connect.hdfs.avro.AvroFormat";
```
ewencp: Are we choosing this as a default for some reason? Does it provide the best support of features, or is it an arbitrary choice?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -25,47 +25,115 @@
...
+
+  public static final String HIVE_INTEGRATION_CONFIG = "hive.integration";
+  private static final String HIVE_INTEGRAtiON_DOC = "Integrate with Hive";
+  public static final boolean HIVE_INTEGRATION_DEFAULT = false;
+
+  public static final String HIVE_DATABASE_CONFIG = "hive.database";
+  private static final String HIVE_DATABASE_DOC = "The database to use when writing to hive.";
+  private static final String HIVE_DATABASE_DEFAULT = "default";
+
+  public static final String AWAIT_TERMINATION_TIMEOUT_CONFIG = "hive.executor.shutdown.timeout.ms";
```
ewencp: This naming is very specific and exposes internal implementation details. What about just having an overall shutdown timeout?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -25,47 +25,115 @@
...
+  public static final String PARTITION_FIELD_CONFIG = "partition.field";
+  private static final String PARTITION_FIELD_DOC = "The field to partition data.";
+  public static final String PARTITION_FIELD_DEFAULT = "";
+
+  public static final String PARTITION_DURATION_CONFIG = "partition.duration.ms";
+  private static final String PARTITION_DURATION_DOC = "The duration of a partition in time based partitoner.";
+  public static final long PARTITION_DURATION_DEFAULT = -1L;
+
+  public static final String PATH_FORMAT_CONFIG = "path.format";
+  private static final String PATH_FORMAT_DOC = "The time format to use.";
```
ewencp: These docs could give a bit more info, e.g. "The time format to use when partitioning with TimeBasedPartitioner."

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -49,14 +50,24 @@ public String version() {
   public void start(Map<String, String> props) {
     try {
       HdfsSinkConnectorConfig connectorConfig = new HdfsSinkConnectorConfig(props);
+      boolean hiveIntegration = connectorConfig.getBoolean(HdfsSinkConnectorConfig.HIVE_INTEGRATION_CONFIG);
+      if (hiveIntegration) {
+        Compatibility compatibility = SchemaUtils.getCompatibility(
+            connectorConfig.getString(HdfsSinkConnectorConfig.SCHEMA_COMPATIBILITY_CONFIG));
+        if (compatibility == Compatibility.NONE) {
```
ewencp: Should all these types of checks just be performed once in the connector? If we check for validity there, then we should never even generate a task configuration that is invalid.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -0,0 +1,561 @@
...
+              }
+            }
+            SinkRecord record = buffer.peek();
+            Schema valueSchema = record.valueSchema();
+            if (shouldRotate()) {
+              nextState();
+            } else if (SchemaUtils.shouldChangeSchema(valueSchema, currentSchema, compatibility)) {
+              currentSchema = valueSchema;
+              if (hiveIntegration) {
+                if (!tableCreated) {
```
ewencp: I'm not sure this logic works. What if the table exists already? The flag `tableCreated` may not be set (this task just got assigned the topic partition but the table exists), and so we'll go into `createHiveTable()`, but I don't see a recovery path that falls back to altering the schema if the table was already created.

src/main/java/io/confluent/connect/hdfs/hive/HiveMetaStore.java
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  private interface ClientAction<R> {
+    R call() throws TException;
+  }
+
+  private <R> R doWithRetry(ClientAction<R> action) throws TException {
+    try {
+      return action.call();
+    } catch (TException e) {
```
ewencp: Is any logging here worthwhile? It'd probably be good to give some indication if the connection is continuously failing.

src/main/java/io/confluent/connect/hdfs/hive/HiveSchemaConverter.java
```diff
@@ -0,0 +1,94 @@
...
+    TYPE_TO_TYPEINFO.put(Type.INT64, TypeInfoFactory.longTypeInfo);
+    TYPE_TO_TYPEINFO.put(Type.FLOAT32, TypeInfoFactory.floatTypeInfo);
+    TYPE_TO_TYPEINFO.put(Type.FLOAT64, TypeInfoFactory.doubleTypeInfo);
+    TYPE_TO_TYPEINFO.put(Type.BYTES, TypeInfoFactory.binaryTypeInfo);
+    TYPE_TO_TYPEINFO.put(Type.STRING, TypeInfoFactory.stringTypeInfo);
+  }
+
+  public static List<FieldSchema> convertSchema(Schema schema) {
+    List<FieldSchema> columns = new ArrayList<>();
+    if (Schema.Type.STRUCT.equals(schema.type())) {
```
ewencp: Can this not be the case? It seems like we're just going to drop anything that's not a struct? That might be a fine limitation, but we a) might want to log something and b) might want to document that somewhere.

ewencp:
@Ishiihara The branch needs a small update to fix up the pom.xml with trunk. More importantly, I can't actually run all the tests. I'm seeing out of memory issues:

```
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 15.936 sec - in io.confluent.connect.hdfs.parquet.HiveIntegrationParquetTest
Exception in thread "main" Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "main"
Exception in thread "Thread-4289" Exception in thread "Thread-4260" Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Thread-4289"
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Thread-4260"
Exception in thread "Thread-4302" Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Thread-4302"
```

But I'm not sure what's causing it because it seems to happen after a set of tests finish and if it was during a test, presumably the log would have shown the other test starting. I've also tried increasing the memory for tests by adding `<argLine>-Xmx2048m</argLine>` to the maven-surefire-plugin arguments, but it's still failing. Any ideas, and have you been seeing this?

ewencp:
@Ishiihara FYI, adding `<forkMode>pertest</forkMode>` to the `maven-surefire-plugin` `<configuration>` section in the `pom.xml` resolves the out of memory issue. Not ideal, but a good enough solution for now.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
Ishiihara: Done.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
Ishiihara: If the DB is not provided, the tables will be created in the Hive default database. We need both configurations.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
Ishiihara: This is simpler as the future is completed in the executor service thread. Will make this change in a following PR.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
Ishiihara: It is better tested. Other than that, it is an arbitrary choice.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
Ishiihara: Done.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
Ishiihara: Done.

src/main/java/io/confluent/connect/hdfs/hive/HiveMetaStore.java
Ishiihara: Done.

src/main/java/io/confluent/connect/hdfs/hive/HiveSchemaConverter.java
Ishiihara: Currently, the connector assumes that the data to write to hdfs is a struct. But it can have nested schemas.

Ishiihara:
@ewencp Just update the PR to address review comments.

ewencp:
@Ishiihara I'd like to rename that one config I just left a comment about, but otherwise LGTM.

PR #14, opened 2015-11-21
System tests

Ishiihara:


pom.xml
```diff
@@ -133,6 +128,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.hadoop</groupId>
```
ewencp: Wow, good catch. Definitely no need to include this in the distribution...

src/assembly/development.xml
```diff
@@ -12,6 +12,16 @@
     <dependencySets>
         <dependencySet>
             <outputDirectory>share/java/kafka-connect-hdfs/</outputDirectory>
+            <useProjectArtifact>true</useProjectArtifact>
+            <useTransitiveFiltering>true</useTransitiveFiltering>
+            <excludes>
```
ewencp: Let's just add a quick comment here to explain why we're making these exclusions. Just needs enough that someone like Michael who comes to look at this for packaging will understand why this is here. Same goes for `package.xml`.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -69,6 +69,8 @@ public void start(Map<String, String> props) {
     } catch (ConfigException e) {
       throw new ConnectException("Couldn't start HdfsSinkConnector due to configuration error.", e);
     } catch (ConnectException e) {
+      log.info("Couldn't start HdfsSinkConnector due to {}",  e.getMessage());
```
ewencp: Let's just use the `log.info("Couldn't start HdfsSinkConnector:",  e);` since that will give the stack trace as well.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -77,13 +79,17 @@ public void start(Map<String, String> props) {
...
+    if (hdfsWriter != null) {
+      hdfsWriter.close();
+    }
   }
 
   @Override
   public void put(Collection<SinkRecord> records) throws ConnectException {
     try {
-      hdfsWriter.write(records);
+      if (hdfsWriter != null) {
```
ewencp: How are we getting here when `hdfsWriter` is null? Shouldn't `start` guarantee that it is not null, or result in an error that causes the connector to stop?

Is this actually a bug in the Kafka Connect runtime where it's invoking this even when `start` fails? Or is this because `ConnectException` in `start` does not throw an exception?

This definitely looks dangerous since we're effectively just dropping data with this code.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -295,15 +297,23 @@ public void close() throws ConnectException {
           commitFile(encodedPartition);
         }
       } catch (IOException e) {
-        log.error("Error rotating {} when closing TopicPartitionWriter.", tempFiles.get(encodedPartition));
+        log.error("Error rotating {} when closing TopicPartitionWriter.",
+                  tempFiles.get(encodedPartition));
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
ewencp: Is all this necessary? Why not just use this form?

```
log.error("Error rotating {} when closing TopicPartitionWriter.",
+                  tempFiles.get(encodedPartition), e);
```

i.e. just add the exception to the end of the original log message, which should dump its message + stacktrace to the log.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -295,15 +297,23 @@ public void close() throws ConnectException {
...
+        log.error(e.toString());
         exceptions.add(e);
       }
-    }
     writers.clear();
     try {
       wal.close();
     } catch (ConnectException e) {
       log.error("Error closing {}.", wal.getLogFile());
+      log.debug(e.getMessage());
```
ewencp: Same here, if we're logging the error, it's probably fine to log the exception that caused it too.

src/main/java/io/confluent/connect/hdfs/avro/AvroRecordWriterProvider.java
```diff
@@ -56,6 +59,7 @@ public String getExtension() {
     return new RecordWriter<SinkRecord>(){
       @Override
       public void write(SinkRecord record) throws IOException {
+        log.debug("Sink record: {}", record.toString());
```
ewencp: I have a follow up patch that removes this anyway, but this results in *very large* log files. That's fine when it's useful for tests, at a low enough log level, and can be isolated from normal logs, but this was a bit excessive :)

ewencp:
@Ishiihara Suggested a few minor tweaks. Otherwise LGTM. Only big issue to address is why we need the null check for `hdfsWriter` in `put()`, which seems like it could result in just reading and dropping data.

Ishiihara:
@ewencp Can you take a look again? I will merge it after this pass.

ewencp:
@Ishiihara LGTM.

PR #15, opened 2015-11-24
Make SchemaSourceTask support rate limiting, use the task ID for the Kafka partition to support multi-partition output, and print info about produced messages to stdout.

ewencp:


ewencp:
@Ishiihara Some tweaks to the SchemaSourceTask to make it work for more than the basic system tests. In particular, for long running tests, we need to be able to rate limit. I went with a simple very approximate approach that is good enough for the tests we want to write immediately. The other change is just a minor generalization to support multiple tasks/partitions better -- you didn't need it for the simple test, but we want it to exercise how the connector handles multiple partitions. In distributed mode we'll want this so we actually have work being distributed across the multiple workers. I also added output specifically to stdout where it can be ignored if not needed, but used for validating output if the exact number of messages produced will not be known, e.g. in hard bounce tests.

Long term, I think we'll want to figure out how to make VerifiableSource/Sink in Kafka generalizable to handle cases like this. This connector looks very similar. A few features are easy to add, such as the max # of messages. The one that I think is tricky is the extra data that we want to include (that we're using for schema validation here). Making it possible to add that generically, especially if any of it needs to vary, will be tricky.

src/main/java/io/confluent/connect/hdfs/tools/SchemaSourceTask.java
```diff
@@ -99,24 +120,25 @@ public void start(Map<String, String> props) {
...
-      log.debug("Source record: {}", srcRecord.toString());
+      SourceRecord srcRecord = new SourceRecord(partition, ccOffset, topic, id, Schema.STRING_SCHEMA, "key", valueSchema, data);
+      System.out.println("{\"task\": " + id + ", \"seqno\": " + seqno + "}");
       List<SourceRecord> result = Arrays.asList(srcRecord);
       seqno++;
       count++;
       return result;
     } else {
+      synchronized (this) {
+        this.wait();
```
Ishiihara: If we reach the max number of messages, the poll will be blocked. This should be OK.

Ishiihara:
LGTM.

src/main/java/io/confluent/connect/hdfs/tools/SchemaSourceTask.java
ewencp: Yeah, the idea is that we should just sleep because there's no work to do (and sleep indefinitely because there will never be any more work to do). I think we actually wanted this before any of the other additions I made. Without it, the task will basically just busy loop by returning empty lists immediately for every iteration.

PR #16, opened 2015-11-24
Check schemas for reference equality to allow for fast path if schemas are reused and no projection is needed.

ewencp:


ewencp:
@Ishiihara Trivial performance optimization -- if we can avoid doing deep equality comparisons of schemas when the schema object is reused, that can have a non-trivial impact on throughput.

Ishiihara:
LGTM

PR #18, opened 2015-11-24
Fix off-by-one error in resetting offsets.

ewencp:
The offset should be one more than the largest offset found in the filenames
since the offset should be set to the next record to be read. Also get rid of
many +1/-1 adjustments to offset, offsets, and beginOffsets in
TopicPartitionWriter to minimize the number of places we have to adjust offset
information by 1.


ewencp:
@Ishiihara An oversight that was caught by the system test I'm working on. The most important parts are that when we look up committed offsets, all of the reset offsets should be + 1 (so the tests were off by one). I also rolled in changed to get rid of a bunch of +1/-1 in maintaining offsets because it left me too confused about what the correct values were...

Ishiihara:
Thanks for this. +1 on this.

PR #20, opened 2015-11-24
Add zero padding of offsets in filenames.

ewencp:


ewencp:
@Ishiihara

Ishiihara:
LGTM

Ishiihara:
@ewencp Should we check negative values and makes sure the value is integer?

ewencp:
@Ishiihara Good point. The type in the config already ensures it's an integer, I added a validator to ensure it's >= 0.

PR #21, opened 2015-11-25
Enable multi-schema data publish

Ishiihara:


src/main/java/io/confluent/connect/hdfs/tools/SchemaSourceTask.java
```diff
@@ -122,6 +133,21 @@ public void start(Map<String, String> props) {
 
       Map<String, Long> ccOffset = Collections.singletonMap(SEQNO_FIELD, seqno);
       SourceRecord srcRecord = new SourceRecord(partition, ccOffset, topic, id, Schema.STRING_SCHEMA, "key", valueSchema, data);
+
```
ewencp: Would be nice to only generate one `Struct` and `SourceRecord` by putting both of these in an `if`/`else` block rather than just the second schema version.

ewencp:
LGTM. Would be nice to clean up the one comment I left, but not a big deal to leave as is.

PR #22, opened 2015-11-26
Drop temp files during onPartitionsRevoked to correctly handle rebalance

ewencp:
Instead of trying to hold onto temp files and continue using them if assigned the same partition after rebalance, just drop the temp files immediately because trying to reuse them requires a much more complicated approach and validation of the offsets since the consumer could have missed some generations of group membership.

As fallout of this change, we also no longer save in-progress work when cleanly shutting down the connector, instead always getting files of the requested size. We also validate offsets of records because of KAFKA-2894 which can result in incorrect delivery of messages to connectors that try to reset offsets during a rebalance operation.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -208,6 +208,9 @@ public class DataWriter {
           tp, storage, writerProvider, partitioner, connectorConfig, context, avroData,
           hiveMetaStore, hive, schemaFileReader, executorService, hiveUpdateFutures);
       topicPartitionWriters.put(tp, topicPartitionWriter);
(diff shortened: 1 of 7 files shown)


ewencp:
@Ishiihara This is the change we discussed previously and makes the distributed system test hard bounce test pass (confluentinc/muckrake#67). There was some test fallout because we don't commit quite the same way anymore, and I also found an issue in the core Kafka Connect framework that we have to work around for now. Also note that I changed when we trigger a file rotation. We used to wait until the write *after* the last entry was written to the file, which meant a) you could potentially not rotate for awhile if no new messages came in for awhile after that and b) tests would have been more annoying to write correctly because you'd constantly have to add extra messages.

Alternatively, @gwenshap if you happen to get to review it first...

gwenshap:
LGTM

PR #24, opened 2015-11-30
Add secure HDFS support

Ishiihara:
@ewencp @gwenshap Added secure HDFS support.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -134,6 +134,27 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-minikdc</artifactId>
+            <version>${hadoop.version}</version>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <!-- exclude this from hadoop minikdc as the minikdc depends on
+                    the apacheds-jdbm1 bundle, which is not available in maven central-->
+                    <groupId>org.apache.directory.jdbm</groupId>
+                    <artifactId>apacheds-jdbm1</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <!-- add this to satisfy the dependency requirement of apacheds-jdbm1-->
+            <groupId>org.apache.directory.jdbm</groupId>
+            <artifactId>apacheds-jdbm1</artifactId>
+            <version>2.0.0-M2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>joda-time</groupId>
             <artifactId>joda-time</artifactId>
             <version>${joda.version}</version>
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -17,6 +17,8 @@ package io.confluent.connect.hdfs;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.security.SecurityUtil;
+import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.errors.ConnectException;
@@ -26,6 +28,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.net.InetAddress;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -41,6 +44,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
+import io.confluent.common.config.ConfigException;
 import io.confluent.connect.avro.AvroData;
 import io.confluent.connect.hdfs.filter.CommittedFileFilter;
 import io.confluent.connect.hdfs.filter.TopicCommittedFileFilter;
@@ -73,11 +77,12 @@ public class DataWriter {
   private HiveUtil hive;
   private Queue<Future> hiveUpdateFutures;
   private boolean hiveIntegration;
+  private Thread ticketRenewThread;
+  private volatile boolean isRunning;
 
   @SuppressWarnings("unchecked")
   public DataWriter(HdfsSinkConnectorConfig connectorConfig, SinkTaskContext context, AvroData avroData) {
     try {
-
       String hadoopHome = connectorConfig.getString(HdfsSinkConnectorConfig.HADOOP_HOME_CONFIG);
       System.setProperty("hadoop.home.dir", hadoopHome);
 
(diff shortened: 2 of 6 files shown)


src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -118,6 +118,18 @@
       + "provide fixed width filenames that can be ordered by simple lexicographic sorting.";
   public static final int FILENAME_OFFSET_ZERO_PAD_WIDTH_DEFAULT = 10;
 
+  public static final String SECURE_HADOOP_CONFIG = "is.secure.hadoop";
+  private static final String SECURE_HADOOP_DOC = "Config indicating whether the hadoop cluster is secure or not";
```
gwenshap: Maybe we should be more specific? "secure" hadoop can mean a bunch of stuff (including HDFS encryption...). 
We only do Kerberos authentication, so I'd like to be clear on this.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -118,6 +118,18 @@
...
+  private static boolean SECURE_HADOOP_DEFAULT = false;
+
+  public static final String HDFS_CONNECTOR_PRINCIPAL_CONFIG = "hdfs.connector.principal";
+  private static final String HDFS_CONNECTOR_PRINCIPAL_DOC =
+      "The principal to use when authenticating the connector with a secure hdfs cluster.";
```
gwenshap: Does the principal need to match an actual OS user? (It does if you create "local" temp files, like MR does). 
If we need it, please document the requirement.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -118,6 +118,18 @@
...
+  public static final String HDFS_CONNECTOR_PRINCIPAL_DEFAULT = "";
+
+  public static final String HDFS_CONNECTOR_KEYTAB_CONFIG = "hdfs.connector.keytab";
+  private static final String HDFS_CONNECTOR_KEYTAB_DOC = "The path to the keytab for the hdfs connector principal.";
```
gwenshap: Which user should have privileges to read this file?

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -58,6 +60,17 @@ public void start(Map<String, String> props) {
           throw new ConfigException("Hive Integration requires schema compatibility to be BACKWARD, FORWARD or FULL");
         }
       }
+
+      boolean secureHadoop = connectorConfig.getBoolean(HdfsSinkConnectorConfig.SECURE_HADOOP_CONFIG);
+      if (secureHadoop) {
+        String principal = connectorConfig.getString(HdfsSinkConnectorConfig.HDFS_CONNECTOR_PRINCIPAL_CONFIG);
+        String keytab = connectorConfig.getString(HdfsSinkConnectorConfig.HDFS_CONNECTOR_KEYTAB_CONFIG);
+
+        UserGroupInformation ugi = UserGroupInformation.loginUserFromKeytabAndReturnUGI(principal,
```
gwenshap: Did you validate that this renews?

gwenshap:
3 lines code addition with 250 lines of tests. Love it :)

pom.xml
```diff
@@ -134,6 +134,27 @@
...
+                    <artifactId>apacheds-jdbm1</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <!-- add this to satisfy the dependency requirement of jdbm-->
+            <groupId>org.apache.directory.server</groupId>
+            <artifactId>apacheds-jdbm</artifactId>
+            <version>2.0.0-M2</version>
+            <scope>test</scope>
```
ewencp: This doesn't seem to be working for me -- it's still complaining about not being able to find the bundle artifact. What version of maven do you have? Mine is `Apache Maven 3.2.5 (12a6b3acb947671f09b81f49094c53f426d8cea1; 2014-12-14T09:29:23-08:00)`. Not sure what else would cause there to be a different since everything else is done via maven plugins.

ewencp: @Ishiihara Actually, I think this just has some typos. In the exclusion, it should be `org.apache.directory.jdbm`, not `org.apache.directory.server`. In the dependency, you need the same change and the artifact ID is missing a 1 at the end, i.e. it should be `apacheds-jdbm1`.

With those changes, it builds and the tests pass.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -118,6 +118,21 @@
       + "provide fixed width filenames that can be ordered by simple lexicographic sorting.";
   public static final int FILENAME_OFFSET_ZERO_PAD_WIDTH_DEFAULT = 10;
 
+  public static final String HDFS_AUTHENTICATE_KERBEROS_CONFIG = "hadoop.authentication.kerberos";
```
ewencp: Seems inconsistent to use the prefix `hadoop.` here but `hdfs.` prefix elsewhere. Is this intentional (especially since the variable name and config key don't match)?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -118,6 +118,21 @@
...
+  private static final String HDFS_AUTHENTICATE_KERBEROS_DOC =
+      "Config indicating whether the hdfs cluster is using Kerberros to perform authentication. ";
```
ewencp: Typo: Kerberros -> Kerberos. Also, we should probably capitalize HDFS throughout these docs.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -58,6 +62,40 @@ public void start(Map<String, String> props) {
...
+        log.info("Login as: " + ugi.getUserName());
+
+        final long renewPeriod = connectorConfig.getLong(HdfsSinkConnectorConfig.KERBEROS_TICKET_RENEW_PERIOD_MS_CONFIG);
+
+        isRunning = true;
+        ticketRenewThread = new Thread(new Runnable() {
+          @Override
+          public void run() {
+            try {
+              while (isRunning) {
```
ewencp: The loop should be inside `synchronized`. If its not, you could get out of the synchronized block, continue the loop after checking `isRunning`, and then wait *after* the other thread set the flag and called `notify`. The thread would then have to wait a full renew period before exiting.

Similarly, you'll want to check the `isRunning` flag when you return from `wait` since you don't want to relogin if you're supposed to be exiting anyway (and you definitely don't want to have to wait for that renewal to complete).

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -118,6 +118,26 @@
...
+  public static final String HDFS_CONNECTOR_KEYTAB_CONFIG = "hdfs.connector.keytab";
+  private static final String HDFS_CONNECTOR_KEYTAB_DOC =
+      "The path to the keytab file for the HDFS connector principal. "
+      + "This keytab file should only be readable by the connector user";
+  public static final String HDFS_CONNECTOR_KEYTAB_DEFAULT = "";
+
+  public static final String KERBEROS_TICKET_RENEW_PERIOD_MS_CONFIG = "kerberos.ticket.renew.period.ms";
+  private static final String KERBEROS_TICKET_RENEW_PERIOD_MS_DOC =
+      "The period in milliseconds to renew the Kerberos ticket.";
+  public static final long KERBEROS_TICKET_RENEW_PERIOD_MS_DEFAULT = 60000 * 60;
```
ewencp: Should this only be an hour? I thought the normal interval would be closer to 1 day? I realize you want to renew before that, but do we really want this to default to only 1 hr?

Ishiihara:
@ewencp @gwenshap Updated the PR to add connect-site.xml. The introduction of connect-site.xml is to provide the flexibility of using different paths of keytab files and connect principals on different hosts. The connect-site.xml needs to present on every node running the hdfs connector.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -90,8 +93,47 @@ public DataWriter(HdfsSinkConnectorConfig connectorConfig, SinkTaskContext conte
...
+              synchronized (DataWriter.this) {
+                while (isRunning) {
+                  wait(renewPeriod);
+                  if (isRunning) {
+                    ugi.reloginFromKeytab();
+                  }
+                }
+              }
+            } catch (IOException e) {
+              log.error("Error renewing the ticket", e);
```
ewencp: Is catching this outside the loop going to cause problems if we, e.g., have a transient connectivity issue? I think it would kill this renewal thread.

I think the exception handling needs to happen in the `while` loop, and we probably need some smaller backoff for retrying in the case of failures (since you probably don't want to wait the entire renewal period if one attempt failed since you need to get it renewed).

src/test/java/io/confluent/connect/hdfs/ConnectSiteXMLTest.java
```diff
@@ -0,0 +1,71 @@
...
+    assertEquals("path_to_keytab", conf.get("connect.hdfs.keytab"));
+    assertEquals("kerberos", conf.get("hadoop.security.authentication"));
+    assertEquals("hdfs/_HOST@EXAMPLE.COM", conf.get("dfs.namenode.kerberos.principal"));
+    FileUtil.fullyDelete(baseDir);
+  }
+
+  private void createTempConnectSiteXMLFile() throws Exception {
+    String template = "<configuration>" +
+                        "<property>" +
+                            "<name>hadoop.security.authentication</name>" +
```
ewencp: Normally this would be in an existing config file, right? Is this just the convenient place to put it during the test to avoid creating another file?

ewencp:
@Ishiihara I had one comment on ensuring we handle errors during refresh properly such that we don't lose the thread, but otherwise this LGTM. But if we can also get @gwenshap to take a look at the approach with connect-site.xml, I think that'd be helpful. She'll be better able to say whether this is something users would be happy with.

Ishiihara:
@ewencp @gwenshap Moved the security related configs back to connector config and move the exception handling of ticket renew thread into while loop.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
gwenshap: Thats fairly normal AFAIK

gwenshap:
LGTM

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
ewencp: @gwenshap Yeah, apparently others will renew a lot more frequently, like on the order of minutes. I don't really get why people would want to do a lot of extra unnecessary work, but if they want to, I won't stop them :)

PR #26, opened 2015-12-02
Add hive.metastore.uris config

Ishiihara:
We add the hive.metasore.uris config to work around a bug in HiveConf.addResource which can set the value of a configuration back to default. This is needed for connector system test to correctly pick up the right metastore. The documentation for the configurations should be improved and I will address them in later PRs. 
@ewencp Should be a quick review.


ewencp:
LGTM

PR #27, opened 2015-12-03
Add timebased file rotation

Ishiihara:
@ewencp Time based file rotation.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -68,6 +68,8 @@ public class TopicPartitionWriter {
   private SinkTaskContext context;
   private int recordCounter;
   private int flushSize;
+  private long rotateIntervalMs;
+  private long lastRotate;
   private RecordWriterProvider writerProvider;
   private Configuration conf;
   private AvroData avroData;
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -370,7 +373,14 @@ private void setState(State state) {
   }
 
   private boolean shouldRotate() {
-    return recordCounter >= flushSize;
+    if (recordCounter >= flushSize) {
+      return true;
+    } else if (rotateIntervalMs <= 0) {
+      return false;
+    } else {
+      long now = System.currentTimeMillis();
```
ewencp: This is going to end up calling `System.currentTimeMillis()` for every record, right? Because to get the record count version to work accurately, this must be called for every record. Should we lift the call out into the parent method, ensure it's only invoked once per batch and pass the value in here?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -497,6 +507,7 @@ private void closeTempFile() throws IOException {
     for (String encodedPartition: tempFiles.keySet()) {
       closeTempFile(encodedPartition);
     }
+    lastRotate = System.currentTimeMillis();
```
ewencp: Is this the right time to do this? I'm wondering if we can end up with closing the file, then running into an issue where it takes quite awhile to actually get the file committed (e.g. we happen to hit HDFS issues then). Then I think the following period wouldn't have been processing messages for as long because we already effectively started it here.

I think we should go entirely one way or the other: if we set `lastRotate` before we've finished the rotation, we should do it immediately when we start (resulting behavior: we rotate every `rotateIntervalMs` regardless of how long committing takes), or we should wait to set `lastRotate` until we're absolutely sure we've finished committing (resulting behavior: process messages for `rotateIntervalMs` before rotating, regardless of how long commits take).

ewencp:
@Ishiihara I think the patch only needs minor tweaks for the code that's there, but I'm also seeing a bunch of unit test failures as a result of this change.

ewencp:
@Ishiihara Fix for tests looks good. This does also resolve the issue with system tests. But what about the issues I raised in the original PR? Once we resolve those, I think we're good to merge both PRs.

Ishiihara:
@ewencp Update the PR to make the tweaks. Do you mind to take another pass. The unit tests passed for me.

ewencp:
LGTM

PR #29, opened 2015-12-04
Move tempfiles to separate directory

Ishiihara:
@ewencp Moves tempfiles to topicsDir/tmp. Unit test passes on my machines.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnecorConstants.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnecorConstants.java
@@ -15,6 +15,11 @@
 package io.confluent.connect.hdfs;
 
 public class HdfsSinkConnecorConstants {
+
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/hdfs/HdfsSinkConnecorConstants.java
```diff
@@ -17,4 +17,5 @@
 public class HdfsSinkConnecorConstants {
   public static final String COMMMITTED_FILENAME_SEPARATOR = "+";
   public static final String COMMMITTED_FILENAME_SEPARATOR_REGEX = "[\\.|\\+]";
+  public static final String TEMPFILE_DIRECTORY = "/tmp/";
```
ewencp: Is this going to be acceptable to everyone as a hard-coded setting? Or does this need to be configurable?

ewencp:
@Ishiihara One question about configurability, but otherwise LGTM.

PR #34, opened 2015-12-08
Fix error with rotate.interval.ms and adding rotation after inactive …

AlexPiermatteo:
If you decide for a big 'flushSize' number on properties but you don't fill completely a file with you data records, this file will not rotate (and remain in tmp state) until is full and during that period it is unusable. I add a timer for close file after an inactivity writing timeout  ( based on 'rotate.interval.ms' property).
At the same time I mantain the original purpose to use that property for additional rotation during records writing.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -70,7 +70,6 @@ public class TopicPartitionWriter {
   private int flushSize;
   private long rotateIntervalMs;
   private long lastRotate;
-  private int startIntervalTmpFilems = 0; //checkpoint ms time for tmp file
   private RecordWriterProvider writerProvider;
   private Configuration conf;
   private AvroData avroData;
@@ -86,6 +85,7 @@ public class TopicPartitionWriter {
   private HdfsSinkConnectorConfig connectorConfig;
   private String extension;
   private final String zeroPadOffsetFormat;
+  private boolean tmpOpened = false;
 
   private final boolean hiveIntegration;
   private String hiveDatabase;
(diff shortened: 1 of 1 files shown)


Ishiihara:
@AlexPiermatteo Thank you for submitting the pull request. It is great that you are interested in contributing to the HDFS connector. 

In our original implementation, we record the timestamp each time we start invoking a file rotation regardless of whether the file rotation is triggered by number of records in the temp reach ``flush.size`` or ``rotate.inverval.ms`` has elapsed. I think your PR is similar to this implementation but you update the timestamp for every record written. However, calling ``System.currentTimeMills` on every record is very expensive and cause performance issues, thus we decided to only check whether we should rotate on every message batch.  

As we only check whether we should perform file rotation each message batch, it is possible that subsequent messages in this batch can trigger file rotation in an earlier time instead of next message batch. However, in normal cases, we should see subsequent messages pretty soon and it is acceptable that we wait a bit more to actually invoke file rotation. 

Also, we use a retry mechanism to handle transient exceptions in ``closeTempFile()``,  ``appendToWAL()`` and ``commitFile()`` to avoid task shutdown. The way that handles file rotation in the PR cat cause the tasks to exit even in case of transient failures.

AlexPiermatteo:
@Ishiihara Thank you for answer me so soon. It's a pleasure try to help on this project. 
I read your answer and I understand your points so yesterday I made a new fix on "TopicPartitionWriter" class.

First of all I think that there was an error during the first writing of a file:
in shouldRotate(now) at the first time lastRotate is 0 so (now - lastRotate) is always > rotateIntervalMs, with this implementation the file rotate only after one message.

Also I mantain the pourpose to close tmp files after an interval of time (using 'rotateIntervalMs' value), even if there are not new messages comming from kafka. I think that it's usefull in order to make query also for data that otherwise in tmp state will not be available.
Now I use your retry mechanism to handle transient exceptions in closeTempFile(), appendToWAL() and commitFile() to avoid task shutdown.
The only thing that I need to do anyway is to calling `System.currentTimeMills on every record.

cotedm:
This one is fixed by #100 so I will close this one now.

PR #42, opened 2016-03-07
Using new config validation

Ishiihara:


Ishiihara:
This depends on https://github.com/apache/kafka/pull/964

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -146,100 +198,223 @@
...
+        .define(CONNECT_HDFS_KEYTAB_CONFIG, Type.STRING, CONNECT_HDFS_KEYTAB_DEFAULT, Importance.HIGH, CONNECT_HDFS_KEYTAB_DOC,
+                SECURITY_GROUP, 3, Width.MEDIUM, CONNECT_HDFS_KEYTAB_DISPLAY, hdfsAuthenticationKerberosDependentsRecommender)
+        .define(HDFS_NAMENODE_PRINCIPAL_CONFIG, Type.STRING, HDFS_NAMENODE_PRINCIPAL_DEFAULT, Importance.HIGH, HDFS_NAMENODE_PRINCIPAL_DOC,
+                SECURITY_GROUP, 4, Width.MEDIUM, HDFS_NAMENODE_PRINCIPAL_DISPLAY, hdfsAuthenticationKerberosDependentsRecommender)
+        .define(KERBEROS_TICKET_RENEW_PERIOD_MS_CONFIG, Type.LONG, KERBEROS_TICKET_RENEW_PERIOD_MS_DEFAULT, Importance.LOW, KERBEROS_TICKET_RENEW_PERIOD_MS_DOC,
+                SECURITY_GROUP, 5, Width.SHORT, KERBEROS_TICKET_RENEW_PERIOD_MS_DISPLAY, hdfsAuthenticationKerberosDependentsRecommender);
+  }
+
+  // Define Schema configuration group
+  static {
```
ewencp: All these static blocks can be combined.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -146,100 +198,223 @@
...
+    config.define(SCHEMA_COMPATIBILITY_CONFIG, Type.STRING, SCHEMA_COMPATIBILITY_DEFAULT, Importance.HIGH, SCHEMA_COMPATIBILITY_DOC, SCHEMA_GROUP, 1, Width.SHORT,
+                  SCHEMA_COMPATIBILITY_DISPLAY, schemaCompatibilityRecommender)
+        .define(SCHEMA_CACHE_SIZE_CONFIG, Type.INT, SCHEMA_CACHE_SIZE_DEFAULT, Importance.LOW, SCHEMA_CACHE_SIZE_DOC, SCHEMA_GROUP, 2, Width.SHORT, SCHEMA_CACHE_SIZE_DISPLAY);
+  }
+
+  // Define Connector configuration group
+  static {
+    config.define(FLUSH_SIZE_CONFIG, Type.INT, Importance.HIGH, FLUSH_SIZE_DOC, CONNECTOR_GROUP, 1, Width.SHORT, FLUSH_SIZE_DISPLAY)
```
ewencp: I'm realizing now that I'm looking at this for this connector -- how do we determine the ordering of the groups?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -14,109 +14,158 @@
...
   private static final String HDFS_URL_DOC =
       "The HDFS connection URL. This configuration has the format of hdfs:://hostname:port and "
       + "specifies the HDFS to export data to.";
+  private static final String HDFS_URL_DISPLAY = "HDFS URL";
 
   public static final String HADOOP_CONF_DIR_CONFIG = "hadoop.conf.dir";
   private static final String HADOOP_CONF_DIR_DOC =
       "The Hadoop configuration directory.";
   public static final String HADOOP_CONF_DIR_DEFAULT = "";
+  private static final String HADOOP_CONF_DIR_DISPLAY = "HADOOP Configuration Directory";
```
ewencp: Hadoop doesn't need to be all caps.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -14,109 +14,158 @@
...
       "Clean shutdown timeout. This makes sure that asynchronous Hive metastore updates are "
       + "completed during connector shutdown.";
   private static final long SHUTDOWN_TIMEOUT_DEFAULT = 3000L;
-
-  public static final String SCHEMA_COMPATIBILITY_CONFIG = "schema.compatibility";
-  private static final String SCHEMA_COMPATIBILITY_DOC =
-      "The schema compatibility rule to use when the connector is observing schema changes. The "
-      + "supported configurations are NONE, BACKWARD, FORWARD and FULL.";
-  private static final String SCHEMA_COMPATIBILITY_DEFAULT = "NONE";
+  private static final String SHUTDOWN_TIMEOUT_DISPLAY = "Shutdown Timeout in Milliseconds";
```
ewencp: Some of these with units might make more sense to list as `"Shutdown Timeout (ms)"` instead.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -146,100 +198,223 @@
...
-      + "This keytab file should only be readable by the connector user.";
-  public static final String CONNECT_HDFS_KEYTAB_DEFAULT = "";
+  // Internal group
+  public static final String STORAGE_CLASS_CONFIG = "storage.class";
+  private static final String STORAGE_CLASS_DOC =
+      "The underlying storage layer. The default is HDFS";
+  public static final String STORAGE_CLASS_DEFAULT = "io.confluent.connect.hdfs.storage.HdfsStorage";
+  private static final String STORAGE_CLASS_DISPLAY = "Storage Class";
+
+  private static final String HDFS_GROUP = "hdfs";
```
ewencp: Should the groups be capitalized? We probably want them to make sense as titles of sections of the UI, and some aren't obvious how to auto-capitalize (like HDFS).

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -146,100 +198,223 @@
...
+  private static final ConfigDef.Recommender hiveIntegrationDependentsRecommender = new BooleanParentRecommender(HIVE_INTEGRATION_CONFIG);
+  private static final ConfigDef.Recommender hdfsAuthenticationKerberosDependentsRecommender = new BooleanParentRecommender(HDFS_AUTHENTICATION_KERBEROS_CONFIG);
+  private static final ConfigDef.Recommender partitionerClassDependentsRecommender = new PartitionerClassDependentsRecommender();
+  private static final ConfigDef.Recommender schemaCompatibilityRecommender = new SchemaCompatibilityRecommender();
+
+  public static ConfigDef config = new ConfigDef();
+
+  // Define HDFS configuration group
+  static {
+    config.define(HDFS_URL_CONFIG, Type.STRING, "", new HdfsUrlValidator(), Importance.HIGH, HDFS_URL_DOC, HDFS_GROUP, 1, Width.MEDIUM, HDFS_URL_DISPLAY)
```
ewencp: Should this actually be an empty string as a default? Or should it not have a default?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -146,100 +198,223 @@
...
+      }
+    }
+
+    @Override
+    public boolean visible(String name, Map<String, String> connectorConfigs) {
+      return true;
+    }
+  }
+  
+  private static class BooleanParentRecommender implements ConfigDef.Recommender {
```
ewencp: This one definitely seems like a very reusable implementation. Should this go somewhere that is easier to share? We can always do this as a refactoring, but worth thinking about what we can reuse.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -146,100 +198,223 @@
...
+    @Override
+    public List<Object> validValues(String name, Map<String, String> connectorConfigs) {
+      return new LinkedList<>();
+    }
+
+    @Override
+    public boolean visible(String name, Map<String, String> connectorConfigs) {
+      String partitionerName = connectorConfigs.get(PARTITIONER_CLASS_CONFIG);
+      if (partitionerName == null) {
+        partitionerName = DefaultPartitioner.class.getName();
```
ewencp: Hmm, so this seems like a pretty significant drawback -- all the logic around parsing that the config classes can already do is not reusable which means connectors will be duplicating lots of code.

Is there any way we can reuse that parsing support? I'm thinking either an AbstractConfig-like class that allows some fields to fail to parse somehow, or maybe just support that in AbstractConfig itself?

src/main/java/io/confluent/connect/hdfs/partitioner/FieldPartitioner.java
```diff
@@ -30,7 +30,7 @@
 import io.confluent.connect.hdfs.HdfsSinkConnectorConfig;
 import io.confluent.connect.hdfs.errors.PartitionException;
 
-public class FieldPartitioner implements Partitioner {
+public final class FieldPartitioner implements Partitioner {
```
ewencp: Why mark this `final`?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnector.java
```diff
@@ -69,4 +73,20 @@ public void start(Map<String, String> props) throws ConnectException {
   public void stop() throws ConnectException {
 
   }
+
+  @Override
+  protected ConfigDef defineConfig() {
+    return HdfsSinkConnectorConfig.getConfig();
```
ewencp: Are we always going to have a class of `AbstractConfig`? Should we think about making that the type that's returned rather than `ConfigDef`. Not really sure which is better, this just looks like in practice it might always be implemented as `return SubclassOfAbstractConfig.getConfig();`.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -14,109 +14,159 @@
...
-      "Number of records written to HDFS before invoking file commits.";
-
-  public static final String ROTATE_INTERVAL_MS_CONFIG = "rotate.interval.ms";
-  private static final String ROTATE_INTERVAL_MS_DOC =
-      "The time interval in milliseconds to invoke file commits. This configuration ensures that "
-      + "file commits are invoked every configured interval. This configuration is useful when data "
-      + "ingestion rate is low and the connector didn't write enough messages to commit files."
-      + "The default value -1 means that this feature is disabled.";
-  private static final long ROTATE_INTERVAL_MS_DEFAULT = -1L;
+  private static final String HADOOP_HOME_DISPLAY = "Haddop home directory";
```
ewencp: Typo - "Haddop"

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -146,100 +199,224 @@
...
+  private static final ConfigDef.Recommender hdfsAuthenticationKerberosDependentsRecommender = new BooleanParentRecommender(HDFS_AUTHENTICATION_KERBEROS_CONFIG);
+  private static final ConfigDef.Recommender partitionerClassDependentsRecommender = new PartitionerClassDependentsRecommender();
+  private static final ConfigDef.Recommender schemaCompatibilityRecommender = new SchemaCompatibilityRecommender();
+
+  private static ConfigDef config = new ConfigDef();
+
+  static {
+
+    // Define HDFS configuration group
+    config.define(HDFS_URL_CONFIG, Type.STRING, "", new HdfsUrlValidator(), Importance.HIGH, HDFS_URL_DOC, HDFS_GROUP, 1, Width.MEDIUM, HDFS_URL_DISPLAY)
```
ewencp: Should this default value be an empty string or `NO_DEFAULT_VALUE`

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -146,100 +199,224 @@
...
+  private static class PartitionerClassDependentsRecommender implements ConfigDef.Recommender {
+
+    @Override
+    public List<Object> validValues(String name, Map<String, String> props) {
+      return new LinkedList<>();
+    }
+
+    @Override
+    public boolean visible(String name, Map<String, String> props) {
+      HdfsSinkConnectorConfig connectorConfig = new HdfsSinkConnectorConfig(props);
```
ewencp: Are we guaranteed this will parse correctly by the time any calls to `visible` are made? This seems a bit wasteful since we end up needing to parse the entire config to check visibility for each config. Maybe this isn't a problem in practice? Or should the parameter actually be different? I think this might come back to the question of whether an `AbstractConfig` subclass should be used instead of `ConfigDef`.

ewencp:
LGTM

PR #46, opened 2016-03-30
Initial implementation

MaxGekk:
This implementation is just initial naive implementation. Any comments are welcome.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,6 +74,11 @@
     <dependencies>
         <dependency>
             <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.11</artifactId>
+            <version>${kafka.version}</version>
+        </dependency>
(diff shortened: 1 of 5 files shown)


DimaD:
Since it went out by accident I think it's still worth to explain what we wanted to achieve.

We are building an in-house data processing system where we use Kafka Connect to write data into HDFS. We are playing with an idea of sending notifications to other parts of data processing system every time new data was successfully committed to HDFS. Naturally we picked Kafka as a transport for those events and are now exploring how exactly we can implement it and if it will cut out for us.

ewencp:
@DimaD This is a great idea -- it might be something that could even be worth supporting at the framework level. Obviously it needs support from the connector, but many connectors will have important events with metadata like this that they may want to publish back into another topic, or will have metrics they'd like to expose.

PR #50, opened 2016-04-18
Remove close of topic partition writers in DataWriter close

Ishiihara:
@ewencp @hachikuji No need to close writers again as we already have it closed in `onPartitionRevoked`.

### DIFF
--- a/src/test/java/io/confluent/connect/hdfs/avro/DataWriterAvroTest.java
+++ b/src/test/java/io/confluent/connect/hdfs/avro/DataWriterAvroTest.java
@@ -294,8 +294,8 @@ public class DataWriterAvroTest extends TestWithMiniDFSCluster {
     Set<TopicPartition> newAssignment = new HashSet<>();
     newAssignment.add(TOPIC_PARTITION);
     newAssignment.add(TOPIC_PARTITION3);
-    hdfsWriter.close(assignment);
 
+    hdfsWriter.close(assignment);
     assignment = newAssignment;
     hdfsWriter.open(newAssignment);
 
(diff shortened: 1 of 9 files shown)


ewencp:
@Ishiihara LGTM.

Doesn't have to happen in this patch, but we should probably update this connector to use the non-deprecated names (i.e. `open` and `close` instead of `onPartitionsAssigned` and `onPartitionsRevoked`.

ewencp:
@Ishiihara Actually, I take that back. It looks like something is wrong with DataWriterAvroTest now:

```
Results :

Tests in error:
  DataWriterAvroTest.testProjectBackWard:415 » FileNotFound File does not exist:...
  DataWriterAvroTest.testProjectForward:533 » FileNotFound File does not exist: ...
  DataWriterAvroTest.testProjectNone:474 » FileNotFound File does not exist: hdf...

Tests run: 45, Failures: 0, Errors: 3, Skipped: 0
```

Ishiihara:
@ewencp Thanks for the review. Fixed the failing tests. PTAL

src/test/java/io/confluent/connect/hdfs/avro/DataWriterAvroTest.java
```diff
@@ -326,7 +331,8 @@ public void testRebalance() throws Exception {
     }
 
     hdfsWriter.write(sinkRecords);
-    hdfsWriter.close();
+    hdfsWriter.close(assignment);
```
ewencp: Shouldn't this `close(newAssignment)` since that was the last assignment opened?

This actually seems pretty concerning -- why isn't `close`-ing the wrong assignment causing an error.

Ishiihara: I think you are referring to the close in testRebalance. In this case, assignment is already refers to the NewAssignment.

ewencp: Ahh, totally makes sense and I just missed the assignment to `assignment`.

ewencp:
LGTM

PR #55, opened 2016-05-12
Removing retry logic from HiveMetaStore to fix the metastore connection bloat

### LINKED ISSUE #54: Too many connections to the hive metastore
We observed that `kafka-connect-hdfs` creates too many connections to the hive metastore, essentially DoS-ing our servers. Ideally, there should never be more than `tasks.max` connections to the metastore since each `HdsSinkTask` creates one `DataWriter`, which creates one `HiveMetaStoreClient`. However, we observed thousands of connections when `tasks.max` was only 200. 

We believe the issue is caused by the retrying logic in `HiveMetaStore`. Specifally, `HiveMetaStore` blindly calls `reconnect` on all `TException`, without ever calling `client.close()`:

```
 74   private <R> R doWithRetry(ClientAction<R> action) throws TException {
 75     try {
 76       return action.call();
 77     } catch (TException e) {
 78       try {
 79           client.reconnect();
 80       } catch (MetaException swallowedException) {
 81         // reconnect failed, throw the original exception
 82         log.warn("Cannot connect to Hive meta store: {}", swallowedException.getMessage());
 83         throw e;
 84       }
 85       return action.call();
 86     }
 87   }

```

Moreover, adding a `client.close()` before `client.reconnect()` won't solve the problem as the underlying layers use a cacheable client wrapper that won't tear down the connection as long as there are users: https://github.com/apache/hive/blob/master/hcatalog/core/src/main/java/org/apache/hive/hcatalog/common/HiveClientCache.java#L394

**Resolution**: From what I understand, the retry logic is unnecessary since the underlying IMetaStoreClient object is actually the `RetryingHiveMetaStoreClient` (as opposed to the vanilla `HiveMetaStoreClient`) which already implements a sane retry logic (calls `reconnect` only after socket timeout). Therefore, we have solved the problem by simply removing the retry logic in `HiveMetaStore.java`

vikeshkhanna:
Fixes #54

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/hive/HiveMetaStore.java
+++ b/src/main/java/io/confluent/connect/hdfs/hive/HiveMetaStore.java
@@ -71,19 +71,15 @@ public class HiveMetaStore {
     R call() throws TException;
   }
 
-  private <R> R doWithRetry(ClientAction<R> action) throws TException {
-    try {
-      return action.call();
-    } catch (TException e) {
-      try {
-          client.reconnect();
-      } catch (MetaException swallowedException) {
-        // reconnect failed, throw the original exception
-        log.warn("Cannot connect to Hive meta store: {}", swallowedException.getMessage());
-        throw e;
-      }
-      return action.call();
-    }
+  private <R> R doAction(ClientAction<R> action) throws TException {
+    // No need to implement retries here. We use RetryingMetaStoreClient
+    // which creates a proxy for a IMetaStoreClient implementation and 
+    // retries calls to it on failure. The retrying client is conscious
+    // of the socket timeout and does not call reconnect on an open connection.
+    // Since HiveMetaStoreClient's reconnect method does not check the status 
+    // of the connection, blind retries may cause a huge spike in the number
+    // of connections to the Hive MetaStore. 
+    return action.call();
   }
 
   public void addPartition(final String database, final String tableName, final String path) throws HiveMetaStoreException {
@@ -99,7 +95,7 @@ public class HiveMetaStore {
     };
 
     try {
-      doWithRetry(addPartition);
+      doAction(addPartition);
     } catch (AlreadyExistsException e) {
       // this is okay
     } catch (InvalidObjectException e) {
@@ -121,7 +117,7 @@ public class HiveMetaStore {
     };
 
     try {
-      doWithRetry(dropPartition);
+      doAction(dropPartition);
     } catch (NoSuchObjectException e) {
       // this is okay
     } catch (InvalidObjectException e) {
@@ -144,7 +140,7 @@ public class HiveMetaStore {
     };
 
     try {
-      doWithRetry(create);
+      doAction(create);
     } catch (AlreadyExistsException e) {
       log.warn("Hive database already exists: {}", database);
     } catch (InvalidObjectException e) {
@@ -167,7 +163,7 @@ public class HiveMetaStore {
     };
 
     try {
-      doWithRetry(drop);
+      doAction(drop);
     } catch (NoSuchObjectException e) {
       // this is okey
     } catch (MetaException e) {
@@ -189,7 +185,7 @@ public class HiveMetaStore {
     createDatabase(table.getDbName());
 
     try {
-      doWithRetry(create);
+      doAction(create);
     } catch (NoSuchObjectException e) {
       throw new HiveMetaStoreException("Hive table not found: " + table.getDbName() + "." + table.getTableName());
     } catch (AlreadyExistsException e) {
@@ -214,7 +210,7 @@ public class HiveMetaStore {
     };
 
     try {
-      doWithRetry(alter);
+      doAction(alter);
     } catch (NoSuchObjectException e) {
       throw new HiveMetaStoreException("Hive table not found: " + table.getDbName() + "." + table.getTableName());
     } catch (InvalidObjectException e) {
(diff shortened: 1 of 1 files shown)


ewencp:
ok to test

ewencp:
@Ishiihara Care to review?

vikeshkhanna:
I am unable to fill the CLA form. It keeps timing out.

vikeshkhanna:
Test failure seems like a build issue (ClassDefNotFoundError):
```
Caused by: java.lang.ClassNotFoundException: org.apache.commons.lang.StringUtils
	at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass(SelfFirstStrategy.java:50)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.unsynchronizedLoadClass(ClassRealm.java:259)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:235)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:227)
```

ewencp:
@vikeshkhanna Sorry about the CLA issue, can you try again now?

vikeshkhanna:
@ewencp Yes, works now. Submitted. Thanks!

Ishiihara:
@ewencp Will take a look.

Ishiihara:
@vikeshkhanna Thank you for the PR and the analysis of the issue. A couple of questions regarding this issue. First of all, is this change fix your problem? I think removing retries definitely reduce the number of connections as we make less number of retries, however, this may not solve the problem entirely. I think the root cause is that the connect tasks are long running and the cached MetaStoreClients are not removed from the cache and thus no connection will be shut down. As any retry will create a a new connection and no connection will be shutdown, eventually, this are still chances that the file descriptors are exhausted. 

One option to fix this issue though, is to set the `hcatalog.hive.client.cache.disabled` configuration in `hive.xml` so that we will not cache the client instance at all, so that any retry will first close the connection and then create a new one.

src/main/java/io/confluent/connect/hdfs/hive/HiveMetaStore.java
```diff
@@ -71,19 +71,10 @@ public HiveMetaStore(Configuration conf, HdfsSinkConnectorConfig connectorConfig
...
-      } catch (MetaException swallowedException) {
-        // reconnect failed, throw the original exception
-        log.warn("Cannot connect to Hive meta store: {}", swallowedException.getMessage());
-        throw e;
-      }
-      return action.call();
-    }
+  private <R> R doAction(ClientAction<R> action) throws TException {
+    // No need to implement retries. We use RetryingHiveMetaStoreClient
+    // which already implements the retrying and reconnect logic
```
Ishiihara: Can you elaborate on what is the retry condition? It seems that two conditions can trigger retry: 1. The connection timeout is exceeded 2. We already retried and we should retry immediately. 

Also, do you know from which Hive version the `RetryingHiveMetaStoreClient` is included? Thanks!

vikeshkhanna: We hit the `AlreadExistsException` (which is a type of `TException`: https://hive.apache.org/javadocs/r0.10.0/api/org/apache/hadoop/hive/metastore/api/AlreadyExistsException.html) very often. This causes the `reconnect` to be called multiple times. Snippet from the logs of our forked code:

```
[2016-05-11 03:31:55,353] [pool-13-thread-1] **WARN [DEBBUG] Action failed. Reconnecting. (io.confluent.connect.hdfs.hive.HiveMetaStore:80)**
AlreadyExistsException(message:Partition already exists:Partition(values:[2016-05-10, 1462849200], dbName:xxxxx, tableName:client_timi
```

We package against the JARs specified in `kafka-connect-hdfs` pom.xml: So I'm guessing the Hive version is 1.2.1? Our deployed Hive version is also 1.2.1.

vikeshkhanna:
@Ishiihara Thanks for looking into this issue!

Our initial investigation (stopped, started and ran Connect in Distributed mode about ~10 times) shows that the number of connections peaked at ~`tasks.max` after the fix. Earlier, at start-up time, the number of connections would go into thousands. I do not have a deep understanding of the caching client or the retrying client, but it seemed clear to me that the `RetryingMetaStoreClient` already implements the `reconnect` logic on failures and is also conscious of the socket timeout (does not call `reconnect` if the connection has been open for less than the timeout). Based solely on the latter, I think the higher level retry was unnecessary

src/main/java/io/confluent/connect/hdfs/hive/HiveMetaStore.java
Ishiihara: That makes sense. I think it worthwhile to remove the retry in the HDFS connector's `HiveMetaStore`.  Would you like to update the comments to elaborate how the `RetryingHiveMetaStoreClient` handles retry?

Ishiihara:
@vikeshkhanna Agreed that the high level retry seems unnecessary. The PR looks good in general and would be nice if you can try to see if setting `hcatalog.hive.client.cache.disabled` solves the issue. Also seems that the `RetryingMetaStoreClient` is added in Hive 0.10.0.

vikeshkhanna:
@Ishiihara I'll try the setting once we resolve a few more blocking operational issues on our side :)

Ishiihara:
@vikeshkhanna Mind to squash the commits? Then I will merge.

Ishiihara:
retest this please

vikeshkhanna:
@Ishiihara Squashed.

Ishiihara:
retest this please

Ishiihara:
LGTM. @vikeshkhanna Thanks for the contribution.

PR #58, opened 2016-05-19
Remove HdfsUrlValidator

Ishiihara:
@ewencp The current validation logic may not work on HDFS when the connect principal have no access to HDFS's root directory.


ewencp:
Do we want to remove it entirely? They're going to need to be able to write *somewhere*. Should we still validate but use `topics.dir` too? (That may be a bunch more work since it wouldn't just be a `Validator` implementation. If it doesn't make sense right now and its better to just remove this to make sure things work ok even if we don't do as much validation, that's fine)

ewencp:
Also, did you want to target this at `master` or `3.x`?

PR #66, opened 2016-05-25
In Hive the schema and partition columns must be disjoint sets

xianzhen:
In Hive the schema and partition columns must be disjoint sets.
btw, method "FieldSchema.equal" doesn't work, cause different FieldSchema.comment. 
There is no comment for fields in schema. when i set "partition.field.name" property in connect config file,partition fields have a comment with empty string "". they don't equal to each other.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java
+++ b/src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java
@@ -23,6 +23,7 @@ import org.apache.hadoop.hive.ql.metadata.HiveException;
 import org.apache.hadoop.hive.ql.metadata.Table;
 import org.apache.kafka.connect.data.Schema;
 
+import java.util.Iterator;
 import java.util.List;
 
 import io.confluent.connect.avro.AvroData;
@@ -69,6 +70,21 @@ public class ParquetHiveUtil extends HiveUtil {
     }
     // convert copycat schema schema to Hive columns
     List<FieldSchema> columns = HiveSchemaConverter.convertSchema(schema);
+    List<FieldSchema> partitionFields = partitioner.partitionFields();
+
+    //In Hive the schema and partition columns must be disjoint sets
+    Iterator<FieldSchema> columnsIter = columns.iterator();
+    for (FieldSchema partitionField : partitionFields) {
+      String partitionFieldName = partitionField.getName();
+
+      while (columnsIter.hasNext()) {
+        FieldSchema column = columnsIter.next();
+        if (column.getName().equals(partitionFieldName)) {
+          columnsIter.remove();
+        }
+      }
+    }
+    
     table.setFields(columns);
     table.setPartCols(partitioner.partitionFields());
     return table;
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -155,6 +155,7 @@ public class TopicPartitionWriter {
         = "%0" +
           connectorConfig.getInt(HdfsSinkConnectorConfig.FILENAME_OFFSET_ZERO_PAD_WIDTH_CONFIG) +
           "d";
+    lastRotate = System.currentTimeMillis();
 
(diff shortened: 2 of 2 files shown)


ewencp:
ok to test

ewencp:
@Ishiihara Care to review?

Ishiihara:
@ewencp Sure.

src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java
```diff
@@ -69,6 +70,21 @@ private Table constructParquetTable(String database, String tableName, Schema sc
     }
     // convert copycat schema schema to Hive columns
     List<FieldSchema> columns = HiveSchemaConverter.convertSchema(schema);
+    List<FieldSchema> partitionFields = partitioner.partitionFields();
+
```
Ishiihara: Sorry I am confused, it seems that Hive didn't check whether there are overlaps between columns and partition columns. Did you see any error if there is overlap? Is this only applies to parquet or to Avro as well?

xianzhen: Thanks for responding. 
Hive didn’t check whether there are overlaps between columns and partition columns. yeap，i describe target table with hive client, everything is fine:hive> 
    > describe singletopic;
OK
time                    string                                      
src_ip                  string                                      
src_port                string                                      
dst_ip                  string                                      
dst_port                string                                      
in_out                  string                                      
partition               string                                      
partition               string                                      

# Partition Information      # col_name              data_type               comment  But, when i use thriftserver, i got this:0: jdbc:hive2://192.168.103.44:10011> describe singletopic;Error: org.apache.spark.sql.AnalysisException: Duplicate column(s) : "partition" found, cannot save to parquet format; (state=,code=0)it looks like the limit of thriftserver.  ------------------------------------------------------------------发件人：Liquan Pei <notifications@github.com>发送时间：2016年5月27日(星期五) 11:06收件人：confluentinc/kafka-connect-hdfs <kafka-connect-hdfs@noreply.github.com>抄　送：先轸0376 <shiq@dtdream.com>; Mention <mention@noreply.github.com>主　题：Re: [confluentinc/kafka-connect-hdfs] In Hive the schema and partition columns must be disjoint sets (#66)
In src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java:> @@ -69,6 +70,21 @@ private Table constructParquetTable(String database, String tableName, Schema sc>      }>      // convert copycat schema schema to Hive columns>      List<FieldSchema> columns = HiveSchemaConverter.convertSchema(schema);> +    List<FieldSchema> partitionFields = partitioner.partitionFields();> +Sorry I am confused, it seems that Hive didn't check whether there are overlaps between columns and partition columns. Did you see any error if there is overlap? Is this only applies to parquet or to Avro as well? —

Ishiihara:
@xianzhen Can you clarify what you want to achieve in this PR. It seems that the Hive schema is not an issue. Do you just want to add `lastRotate`? If that is the case, I suggest that to only include the second commit in the PR and change the title to a more relevant name.

Ishiihara:
@xianzhen It seems that #87 also handles the last rotate. Would you mind to close the PR?

shrijeet:
I am curious why this was closed. I faced similar issue (connect created a hive table which duplicate columns as reported by original poster). How does #87 fixes the table creation logic? Atleast in my case it surfaces when I use 'partitioner.class=io.confluent.connect.hdfs.partitioner.FieldPartitioner'. @Ishiihara thoughts?

Ishiihara:
@shrijeet I was referring to #87 regarding the second commit of this PR which is related to scheduled file rotate.  For the duplicate column issue, can Hive or Spark can query the table when there are duplicate columns? It seems that Hive CLI is fine but not Spark. If that is the case, we need to have a fix.

shrijeet:
@Ishiihara that is correct - hive cli is fine but thriftserver isn't. Thriftserver is what spark uses as well to interact with hive.

xianzhen:
@Ishiihara If you use "partition.field.name" to create partition columns, you will lost the "comment" of this filed. Take this into account. Thanks.

PR #67, opened 2016-05-25
clean up hive metastore artifacts from testing

coughman:


ewencp:
LGTM. And thanks for patches like this, its nice to keep things tidy!

PR #80, opened 2016-06-21
Close file when partition changes

### LINKED ISSUE #71: File not written when partition changes
I have setup with `flush.size=100000` and `TimeBasedPartitioner` with `path.format=YYYY/MM/dd`.

For topics with small amount of messages this cause strange behavior. When partitioner found that partition should be changed it opens new file, but did not close the old one. So, I have not data until `flush.size` will be reached and unable to run some data dependent tasks like ETL and so.

Expected behavior for me is to close file when new partition started. What do you think?

13h3r:
Fixes #71

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -85,6 +85,7 @@ public class TopicPartitionWriter {
   private HdfsSinkConnectorConfig connectorConfig;
   private String extension;
   private final String zeroPadOffsetFormat;
+  private String currentPartition;
 
   private final boolean hiveIntegration;
   private String hiveDatabase;
@@ -261,15 +262,25 @@ public class TopicPartitionWriter {
               }
             } else {
               SinkRecord projectedRecord = SchemaUtils.project(record, currentSchema, compatibility);
-              writeRecord(projectedRecord);
-              buffer.poll();
-              if (shouldRotate(now)) {
+              String partition = partitioner.encodePartition(projectedRecord);
+              if (currentPartition != null && !currentPartition.equals(partition)) {
+                currentPartition = partition;
                 log.info("Starting commit and rotation for topic partition {} with start offsets {}"
-                         + " and end offsets {}", tp, startOffsets, offsets);
+                        + " and end offsets {}", tp, startOffsets, offsets);
                 nextState();
-                // Fall through and try to rotate immediately
               } else {
-                break;
+                currentPartition = partition;
+
+                writeRecord(projectedRecord, partition);
+                buffer.poll();
+                if (shouldRotate(now)) {
+                  log.info("Starting commit and rotation for topic partition {} with start offsets {}"
+                          + " and end offsets {}", tp, startOffsets, offsets);
+                  nextState();
+                  // Fall through and try to rotate immediately
+                } else {
+                  break;
+                }
               }
             }
           case SHOULD_ROTATE:
@@ -459,7 +470,7 @@ public class TopicPartitionWriter {
     }
   }
 
-  private void writeRecord(SinkRecord record) throws IOException {
+  private void writeRecord(SinkRecord record, String encodedPartition) throws IOException {
     long expectedOffset = offset + recordCounter;
     if (offset == -1) {
       offset = record.kafkaOffset();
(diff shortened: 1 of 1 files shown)


13h3r:
[clabot:check]

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -261,15 +262,25 @@ public void write() {
               }
             } else {
               SinkRecord projectedRecord = SchemaUtils.project(record, currentSchema, compatibility);
-              writeRecord(projectedRecord);
-              buffer.poll();
-              if (shouldRotate(now)) {
+              String partition = partitioner.encodePartition(projectedRecord);
+              if (currentPartition != null && !currentPartition.equals(partition)) {
+                currentPartition = partition;
```
ewencp: I'm not sure I understand the goal of this logic. This task might be processing data for a lot of partitions, right? Won't this mean that if the framework delivers data for 3 topic partitions (say that it is conveniently split into 3 sets for each topic partition), we'll close both the first topic partition's and second topic partition's files by the time we get to the third?

Doesn't this violate the semantics that you want from the connector, which is that it tries to keep files open until enough data has been delivered into them? This is *really* important for HDFS since it often won't deal well with files that are, e.g., < 1GB. Of course it *can* handle smaller files, but to optimize performance, its better if you use larger files. The way we currently handle rolling files does this correctly. As far as I can tell, this change will result in a *ton* of tiny files given the way data is fetched from Kafka while your task is handling multiple partitions.

In particular, I think you are addressing the case of `TimeBasedPartitioner` specifically, where you are *also* assuming that timestamped data is delivered in order. That happens to be the case at the moment because of when this feature was added. But if `TimeBasedPartitioner` uses built-in Kafka timestamps, then its possible even within a topic partition, or across topic partitions (which this task might be consuming multiple of) we may see out of order data. For a general partitioner, the policy you're trying to use here doesn't make sense -- if you use `FieldPartitioner` there could be arbitrary ordering of records destined for different files.

I think this probably requires a slightly more complicated solution. One possibility would be to have the partitioner understand which partition files have "expired" and allow this code to handle that. I think that's actually a pretty major change as the `nextState()` call may depend on the current record. I think it may also require thinking through how we handle multi-file commit a bit more thoroughly.

@Ishiihara thoughts? You probably remember the semantics maintained across all these classes better than I do.

13h3r: Thanks for reply! You are right, this will no work with `FieldPartitioner` and it assumes that timestamps will be in order.

The idea here is to provide data for external ETL pipeline in predictable time even if number of messages in a chunk is not reached. Do you have any idea how can I achieve this with kafka connect right now?

Thanks!

ewencp: Does the `rotate.interval.ms` setting not work for you? The two conditions currently checked in `shouldRotate` are whether there are enough messages or if enough time has elapsed.

13h3r: I would like to cut the file exactly at the end of hour. Is it possible to adjust `rotate.interval.ms` to perform rotation with adjusted offset? Something like `getPartition` does for time adjustment?

ewencp: I see, that is not possible at the moment. I think you'd want something like `rotate.schedule.ms` that works a bit more like a cron specification than `rotate.interval.ms` which is just the interval but says nothing about starting time/alignment.

13h3r: @ewencp what about cron expression? The problem with `rotate.schedule.ms` is values that does not fit well. Like 126764ms.

ewencp: Yeah @13h3r that's what I was thinking. Although actually using a cron expression doesn't seem great since people often struggle to get the expression they want correct. An alternative would be to still use `rotate.schedule.ms` but add another config that does something like control alignment, e.g. it forces the intervals to start at 00:00:00 of each day.

One thing to be careful of here is also time synchronization between different systems. This actually may be an issue with the existing code as well... If you move a task, you don't want it to overwrite a file that some other node already wrote just because the clocks of the two systems are not perfectly synchronized.

13h3r:
@ewencp what about keep using `rotate.interval.ms` but introduce option to enable time adjustment? Something like `rotate.interval.adjustment = true`?

ewencp:
@13h3r Yeah, that's basically what I was getting at.

PR #86, opened 2016-07-05
Init lastRotate to system time

xianzhen:
Init lastRotate to system time in constructor. The default value of lastRotate is 0, When the first record comes, 'now - lastRotate' is absolutely bigger than rotateIntervalMs. So, kafka connect will write a file with one record into hdfs.


13h3r:
IMO it is better to init last recovery time after recovery completed. I did it in #87

ewencp:
Yeah, I think @13h3r's suggestion makes sense. If it is already in #87, which I think is converging on a mergeable version, we may be able to skip this PR.

PR #87, opened 2016-07-07
Scheduled rotation implementation

13h3r:


13h3r:
Based on discussions of #80 and fixes #71

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -290,7 +300,8 @@
 
     // Define Connector configuration group
     config.define(FLUSH_SIZE_CONFIG, Type.INT, Importance.HIGH, FLUSH_SIZE_DOC, CONNECTOR_GROUP, 1, Width.SHORT, FLUSH_SIZE_DISPLAY)
-        .define(ROTATE_INTERVAL_MS_CONFIG, Type.LONG, ROTATE_INTERVAL_MS_DEFAULT,Importance.HIGH, ROTATE_INTERVAL_MS_DOC, CONNECTOR_GROUP, 2, Width.SHORT, ROTATE_INTERVAL_MS_DISPLAY)
+        .define(ROTATE_INTERVAL_MS_CONFIG, Type.LONG, ROTATE_INTERVAL_MS_DEFAULT, Importance.HIGH, ROTATE_INTERVAL_MS_DOC, CONNECTOR_GROUP, 2, Width.SHORT, ROTATE_INTERVAL_MS_DISPLAY)
+        .define(ROTATE_SCHEDULE_MS_CONFIG, Type.LONG, ROTATE_SCHEDULE_MS_DEFAULT, Importance.HIGH, ROTATE_SCHEDULE_MS_DOC, CONNECTOR_GROUP, 2, Width.SHORT, ROTATE_SCHEDULE_MS_DISPLAY)
```
Ishiihara: The order in group needs to be adjusted.  This is used to order the configs inside a group. Also, should the importance of this config to be high? I think we can use medium and the reasoning is that the connector still function properly if this config is not provided.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -148,6 +148,16 @@
   private static final long ROTATE_INTERVAL_MS_DEFAULT = -1L;
   private static final String ROTATE_INTERVAL_MS_DISPLAY = "Rotate Interval (ms)";
 
+  public static final String ROTATE_SCHEDULE_MS_CONFIG = "rotate.schedule.ms";
+  private static final String ROTATE_SCHEDULE_MS_DOC =
```
Ishiihara: Maybe use `rotate.schedule.interval.ms'?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -165,6 +171,15 @@ public TopicPartitionWriter(
       this.hiveUpdateFutures = hiveUpdateFutures;
       hivePartitions = new HashSet<>();
     }
+
+    String timeZoneString = (String) connectorConfig.getString(HdfsSinkConnectorConfig.TIMEZONE_CONFIG);
```
Ishiihara: Should we throw the exception only when time config is definitely required. We can check that if the scheduled file rotate is enabled to determine whether we need the timezone config or not. Another option is to introduce another configuration indicating that server timezone should be used instead of  UTC.

src/main/java/io/confluent/connect/hdfs/DateTimeUtils.java
```diff
@@ -0,0 +1,13 @@
+package io.confluent.connect.hdfs;
+
+import org.joda.time.DateTimeZone;
+
+public class DateTimeUtils {
+    private static long DAY_IN_MS = 24 * 60 * 60 * 1000;
+    public static long getNextTimeAdjustedByDay(long currentTimeMs, long periodMs, DateTimeZone timeZone) {
```
Ishiihara: Would be nice to add some documentation on what this utils is trying to achieve.

Ishiihara:
@13h3r Thanks for the contribution. I left a couple of comments. We should think about the default behavior when adding the scheduled file rotate. Should we default to round to the next period of the server timezone?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
13h3r: ok

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
13h3r: agreed

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
13h3r: Agreed on checking only only of scheduled rotation enabled. I want to keep timezone explicit to avoid errors during environment changes

src/main/java/io/confluent/connect/hdfs/DateTimeUtils.java
13h3r: ok

13h3r:
@Ishiihara Thanks for comments. I do not like implicit timezone here, because different servers may have different timezones which causes different behavior for same configuration. I would like to explicitly set timezone for such cases.

13h3r:
Also, do you have CI to check all test?

ewencp:
ok to test

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -165,6 +172,15 @@ public TopicPartitionWriter(
       this.hiveUpdateFutures = hiveUpdateFutures;
       hivePartitions = new HashSet<>();
     }
+
+    if(rotateScheduleIntervalMs > 0) {
+      String timeZoneString = (String) connectorConfig.getString(HdfsSinkConnectorConfig.TIMEZONE_CONFIG);
```
Ishiihara: Type conversion is not needed here.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -165,6 +172,15 @@ public TopicPartitionWriter(
...
+      if (timeZoneString.equals("")) {
+        throw new ConfigException(HdfsSinkConnectorConfig.TIMEZONE_CONFIG,
+                timeZoneString, "Timezone cannot be empty.");
```
Ishiihara: The error message should be more explicit. How about "timezone cannot be empty when using scheduled file rotation."

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -165,6 +172,15 @@ public TopicPartitionWriter(
       this.hiveUpdateFutures = hiveUpdateFutures;
       hivePartitions = new HashSet<>();
     }
+
+    if(rotateScheduleIntervalMs > 0) {
```
Ishiihara: I think it's probably better to move this check to `DataWriter` or even to `HdfsSinkTask` so that we can fail early.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -148,6 +148,16 @@
   private static final long ROTATE_INTERVAL_MS_DEFAULT = -1L;
   private static final String ROTATE_INTERVAL_MS_DISPLAY = "Rotate Interval (ms)";
 
+  public static final String ROTATE_SCHEDULE_INTERVAL_MS_CONFIG = "rotate.schedule.interval.ms";
+  private static final String ROTATE_SCHEDULE_INTERVAL_MS_DOC =
+          "The time interval in milliseconds to periodically invoke file commits. This configuration ensures that "
+          + "file commits are invoked every configured interval. Time of commit will be adjusted to 00:00 of selected timezone. "
```
Ishiihara: Time of commit will be adjusted to 00:00 is a bit confused. Do you mean that the commit time is rounded to the starting time of the next period?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -148,6 +148,16 @@
...
+          + "Commit will be performed at scheduled time regardless previous commit time or number of messages. "
+          + "This configuration is useful when you have to commit your data based on current server time, like at the beginning of every hour. "
+          + "The default value -1 means that this feature is disabled.";
+  private static final long ROTATE_SCHEDULE_INTERVAL_MS_DEFAULT = -1L;
+  private static final String ROTATE_SCHEDULE_INTERVAL_MS_DISPLAY = "Rotate Schedule (ms)";
```
Ishiihara: Rotate Schedule (ms) -> Rotate Schedule Interval (ms)

src/main/java/io/confluent/connect/hdfs/DateTimeUtils.java
```diff
@@ -0,0 +1,22 @@
...
+public class DateTimeUtils {
+    private static long DAY_IN_MS = 24 * 60 * 60 * 1000;
+
+    /**
+     * Calculates next period of periodMs after currentTimeMs starting from midnight in given timeZone.
+     * If the next period is in next day then 12am of next day will be returned
+     * @param currentTimeMs time to calculate at
+     * @param periodMs period in ms
+     * @param timeZone timezone to get midnight time
+     * @return timestamp in ms
```
Ishiihara: Which time zone? It seems that the returned timestamp is in UTC.

src/main/java/io/confluent/connect/hdfs/DateTimeUtils.java
```diff
@@ -0,0 +1,22 @@
+package io.confluent.connect.hdfs;
+
+import org.joda.time.DateTimeZone;
+
+public class DateTimeUtils {
+    private static long DAY_IN_MS = 24 * 60 * 60 * 1000;
```
Ishiihara: final

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -218,6 +234,19 @@ public boolean recover() {
     return true;
   }
 
+  private void updateRotationTimers() {
+    lastRotate = System.currentTimeMillis();
+    if(log.isDebugEnabled() && rotateIntervalMs > 0) {
+      log.debug("Update last rotation timer. Next rotation for {} will be in {}ms", tp, rotateIntervalMs);
+    }
+    if (rotateScheduleIntervalMs > 0) {
+      nextScheduledRotate = DateTimeUtils.getNextTimeAdjustedByDay(System.currentTimeMillis(), rotateScheduleIntervalMs, timeZone);
```
Ishiihara: I think we can use the `lastRotate` here to avoid an additional system call.

src/test/java/io/confluent/connect/hdfs/DateTimeUtilsTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.joda.time.DateTime;
+import org.junit.Test;
+
+import static io.confluent.connect.hdfs.DateTimeUtils.getNextTimeAdjustedByDay;
+import static org.junit.Assert.assertEquals;
+
+
+public class DateTimeUtilsTest {
+    DateTime midnight = DateTime.now().withTimeAtStartOfDay();
```
Ishiihara: private final

src/test/java/io/confluent/connect/hdfs/DateTimeUtilsTest.java
```diff
@@ -0,0 +1,44 @@
+package io.confluent.connect.hdfs;
+
+import org.joda.time.DateTime;
+import org.junit.Test;
+
+import static io.confluent.connect.hdfs.DateTimeUtils.getNextTimeAdjustedByDay;
```
Ishiihara: Let's not statically import `getNextTimeAdjustedByDay`

src/test/java/io/confluent/connect/hdfs/DateTimeUtilsTest.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.junit.Assert.assertEquals;
+
+
+public class DateTimeUtilsTest {
+    DateTime midnight = DateTime.now().withTimeAtStartOfDay();
+
+    private DateTime calc(DateTime current, long periodMs) {
+        return new DateTime(getNextTimeAdjustedByDay(
```
Ishiihara: `getNextTimeAdjustedByDay` ->  `DateTimeUtils.getNextTimeAdjustedByDay`

Ishiihara:
@13h3r Left a couple of more comments. Most of them a simple clean ups. I think we should move the time config check either to `DataWriter` or `HdfsSinkTask` to fail early.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
13h3r: ok

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
13h3r: Moving check into another class and use check result is very fragile for me. We could pass this properties to constructor, but `TopicPartitionWritter` designed to get his properties from config. For me it is ok, to keep this check in `TopicPartitionWritter`.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
13h3r: I mean that in case of interval that does not fit to a day (like 7 hours for example) it will be forced to fit into a day. For 7 hours interval it will trigger commit at 00:00, 07:00, 14:00, 21:00 and 00:00 of next day. Any ideas how to explain it in one sentence?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
13h3r: ok

src/main/java/io/confluent/connect/hdfs/DateTimeUtils.java
13h3r: This timezone used to determine midnight time to adjust rotation time to

src/main/java/io/confluent/connect/hdfs/DateTimeUtils.java
13h3r: ok

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
13h3r: ok

src/test/java/io/confluent/connect/hdfs/DateTimeUtilsTest.java
13h3r: ok

src/test/java/io/confluent/connect/hdfs/DateTimeUtilsTest.java
13h3r: ok

src/test/java/io/confluent/connect/hdfs/DateTimeUtilsTest.java
13h3r: ok

13h3r:
@Ishiihara fixed most of the comments. 

I am not sure about moving config check from `TopicPartitionWriter` because I do not want to separate config checking from config using. We may forget to correct checks in other files when we change `TopicPartitionWriter`.

Ishiihara:
@13h3r I think we should move the config check to `HdfsSinkTask` and here is the reasoning: when the sink task is created and we already know that if the timezone is not provided, the scheduled file rotate is not working properly. We don't want to kill the task after we have synced with Hive, created a thread pool and created a lot of objects. Those options are expensive and not useful if we can not proceed.

13h3r:
Done

13h3r:
@Ishiihara anything else to merge this?

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -59,6 +60,17 @@ public void start(Map<String, String> props) {
...
       }
+
+      //check that timezone it setup correctly in case of scheduled rotation
+      if(connectorConfig.getLong(HdfsSinkConnectorConfig.ROTATE_SCHEDULE_INTERVAL_MS_CONFIG) > 0) {
+        String timeZoneString = connectorConfig.getString(HdfsSinkConnectorConfig.TIMEZONE_CONFIG);
+        if (timeZoneString.equals("")) {
+          throw new ConfigException(HdfsSinkConnectorConfig.TIMEZONE_CONFIG,
+                  timeZoneString, "Timezone cannot be empty when using scheduled file rotation.");
+        }
+        DateTimeZone.forID(timeZoneString);
```
Ishiihara: This is not needed here, right?

Ishiihara:
@13h3r Thanks for the fixes. Left one comment. Also, can you squash the commits after fixing the comments? I should be able to have it merged after squash.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
13h3r: What if we set timezone to something invalid like `asdlkfjak`?

13h3r:
Done

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
Ishiihara: In this case, can we catch the exception and wrap it into a `ConfigException`?

Ishiihara:
@13h3r I am going to merge this PR. Would you like to file a follow up PR to add documentation for this feature?

PR #92, opened 2016-07-22
Start a new WAL file after `truncate` instead of appending to log.1

jingw:
Currently, the code continues appending to log.1, so the next recovery invocation doesn't know about all those WAL entries.

I verified that the regression test fails without the patch.


Ishiihara:
ok to test

Ishiihara:
LGTM. Thanks for the contribution.

PR #95, opened 2016-07-25
adding an option to exclude the topic name from the output folder in HDFS

stheppi:
The default behaviour for the sink is to append the topic name to the "topics.dir". While that is valid scenario, there are cases when multiple topics source the same Hive external table. Therefore by adding the topic name will not make the data visible. Therefore I have added an option to instruct the sink to not create those HDFS directories as it stores the data.

Scenario:
topic1 and topic2 should both store data in hdfs://.../MyDatabase/TableA
TableA is partitioned by eventId

Current outcome:
 These folders are created:
 hdfs://.../MyDatabase/TableA/topic1/eventId=??
 hdfs://.../MyDatabase/TableA/topic2/eventId=??

if I query my table the data won't be visible

Setting topic.name.include = false
Outcome:
 hdfs://.../MyDatabase/TableA/eventId=??
 hdfs://.../MyDatabase/TableA/eventId=??

By default topic.name.include = true so the current behaviour doesn't change

the Parquet library is using JUL and therefore the output is quite verbose. I create an utility class ParquetLogMute to remove the bindings and thus avoid the noise in the logs with the info of writing the values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Kafka Connect HDFS Connector
 
-kafka-connect-hdfs is a [Kafka Connector](http://kafka.apache.org/090/documentation.html#connect)
+kafka-connect-hdfs is a [Kafka Connector](http://kafka.apache.org/documentation.html#connect)
 for copying data between Kafka and Hadoop HDFS.
 
 # Development
--- a/pom.xml
+++ b/pom.xml
@@ -23,7 +23,7 @@
     <groupId>io.confluent</groupId>
     <artifactId>kafka-connect-hdfs</artifactId>
     <packaging>jar</packaging>
-    <version>3.1.0-SNAPSHOT</version>
+    <version>3.0.0</version>
     <name>kafka-connect-hdfs</name>
     <organization>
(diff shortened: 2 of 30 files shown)


ewencp:
@stheppi I'm curious -- this seems pretty extensive, but the key change is isolated to one set of classes -- `Partitioners`. Did you consider an approach that focuses only on those classes? In particular, they have a `configure()` method, so they can be parameterized to help limit the scope and impact of the changes (admittedly with the drawback that we don't get the setting into the documentation automatically).

The drawback with the approach you've taken is that it affects all Partitioners, even for custom implementations where people don't care about the topic name (or the flag you're passing in). I'm not sure what information is forwarded to `configure()` today, but forwarding all the original configs from the task might be enough to support this with a much smaller patch and without assuming all Partitioners will have the same set of flags.

stheppi:
@ewencp ,
 I missed the approach on Partitioners. Indeed it makes a lot more sense.Please reject the PR. I will re-implement the logic as per your suggestion.

PR #96, opened 2016-07-26
Check empty WAL during recovery

Ishiihara:


Ishiihara:
@ewencp Ping for review. Thanks!

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -102,6 +103,20 @@ public void apply() throws ConnectException {
...
+      // This can cause the recovery to persistently fail.
+      FileStatus[] statuses = storage.listStatus(logFile);
+
+      if (statuses.length != 1) {
+        throw new ConnectException("Expected only one WAL file: " + logFile);
+      }
+
+      if (statuses[0].getLen() == 0) {
+        log.warn("Skip empty WAL file: " + logFile);
+        return;
```
ewencp: Wouldn't we still want to acquire the lease here? Shouldn't we be acquiring if for no other reason than to make sure that we *can* at this point? Otherwise, even if its zero bytes now, couldn't something else continue write entries into the WAL.

ewencp:
@Ishiihara looks good other than my one comment. Does this cover CC-100?

Ishiihara:
@ewencp Can you take another look? Thanks!

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -102,7 +102,21 @@ public void apply() throws ConnectException {
       if (!storage.exists(logFile)) {
         return;
       }
+
```
ewencp: Just to check -- is the assumption here that even if we `apply()` and effectively do nothing, even if another process exists and then subsequently writes to the WAL, not acquiring the WAL lease is ok because we will acquire it later when writing and see the conflict?

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -63,7 +66,8 @@
 
   private WALFile() {}
 
-  public static Writer createWriter(Configuration conf, Writer.Option... opts) throws IOException {
+  public static Writer
+  createWriter(Configuration conf, Writer.Option... opts) throws IOException {
```
ewencp: nit: formatting -- shouldn't end up on separate line

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -208,20 +212,29 @@ public static Option blockSize(long value) {
...
-            && fs.exists(p)) {
-          // Read the file and verify header details
-          try (WALFile.Reader reader =
-                   new WALFile.Reader(conf, WALFile.Reader.file(p), new Reader.OnlyHeaderOption())){
-            if (reader.getVersion() != VERSION[3]) {
-              throw new VersionMismatchException(VERSION[3], reader.getVersion());
+        try {
+          out = fs.create(p, false, bufferSize, replication, blockSize);
+        } catch (FileAlreadyExistsException e) {
+          if (appendIfExistsOption != null && appendIfExistsOption.getValue() && fs.exists(p)) {
```
ewencp: Do we really need to check again if it exists? We already know it existed a moment ago, and even with the `fs.exists(p)` call there is still a race for the `fs.append()` call.

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -208,20 +212,29 @@ public static Option blockSize(long value) {
...
+                throw new VersionMismatchException(VERSION[3], reader.getVersion());
+              }
+              sync = reader.getSync();
+              this.appendMode = true;
+            } catch (EOFException eof) {
+              FileStatus status = fs.getFileStatus(p);
+              if (status.getLen() == 0) {
+                this.appendMode = false;
+              } else {
+                throw eof;
```
ewencp: So for now we're assuming that any incomplete header that isn't an empty file will cause the connector to stop running? Is there a reason an empty file should be treated differently from an incomplete header? Wouldn't the latter imply the same problem as the former, i.e. that the previous process writing the WAL had crashed and left it in a corrupt state where it needs to be truncated/deleted to discard it?

kkonstantine:
@C0urante double checking: I believe this PR is superseded by your recently merged PR: https://github.com/confluentinc/kafka-connect-hdfs/pull/348

Would you mind confirming? We can close this one if that's the case.

C0urante:
@kkonstantine correct, we can close this one.

PR #99, opened 2016-08-06
adding an option to exclude the topic name from the output folder in HDFS

stheppi:
The default behaviour for the sink is to append the topic name to the "topics.dir". While that is valid scenario, there are cases when multiple topics source the same Hive external table. Therefore by adding the topic name will not make the data visible. Therefore I have added an option to instruct the sink to not create those HDFS directories as it stores the data.
Scenario:
topic1 and topic2 should both store data in hdfs://.../MyDatabase/TableA
TableA is partitioned by eventId
Current outcome:
These folders are created:
hdfs://.../MyDatabase/TableA/topic1/eventId=??
hdfs://.../MyDatabase/TableA/topic2/eventId=??
if I query my table the data won't be visible
Setting topic.name.include = false
Outcome:
hdfs://.../MyDatabase/TableA/eventId=??
hdfs://.../MyDatabase/TableA/eventId=??
By default topic.name.include = true so the current behaviour doesn't change
the Parquet library is using JUL and therefore the output is quite verbose. I create an utility class ParquetLogMute to remove the bindings and thus avoid the noise in the logs with the info of writing the values.

The code has been changed from the prev PR to follow Ewen's advise

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Kafka Connect HDFS Connector
 
-kafka-connect-hdfs is a [Kafka Connector](http://kafka.apache.org/090/documentation.html#connect)
+kafka-connect-hdfs is a [Kafka Connector](http://kafka.apache.org/documentation.html#connect)
 for copying data between Kafka and Hadoop HDFS.
 
 # Development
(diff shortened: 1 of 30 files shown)


andrewegel:
This is causing our Jenkins CI to break due to the "unknown repository" - This needs to be closed to get around that. If this change is still required, please re-do it in a fork or branch.

Issue #79, opened 2016-06-18
The last few records in tempFiles can't flush to HDFS when stream in kafka in paused


lakeofsand:
The key part(TopicPartitionerWriter.java):

**//when buffer.poll get the last record,but 'shouldRotate(now)' is false_,no subsequent action will occur**

`
while(!buffer.isEmpty()) {

...

SinkRecord projectedRecord = SchemaUtils.project(record, currentSchema, compatibility);
              writeRecord(projectedRecord);
              buffer.poll();
              if (shouldRotate(now)) {
                log.info("Starting commit and rotation for topic partition {} with start offsets {}"
                         + " and end offsets {}", tp, startOffsets, offsets);
                nextState();
                // Fall through and try to rotate immediately
              } else {`

xianzhen:
We are encountering the same issue.

Ishiihara:
That is the current design. I think the workaround here is to set `flush.interval.ms` to a reasonable value? Also, after #87 is merged, the scheduled flush can also be used as a workaround.

skyahead:
We faced the same issue, and we tried setting flush.interval.ms but it did not help if the stream is paused. The reason is shouldRotate() is only checked inside the while(!buffer.isEmpty()) loop. But when the stream is paused, the buffer is also empty.

In our case, our stream gets data every 5 minute and we flush.size is 50K in size. So, we always have 0 to 50K records kept in temp files until new data is coming. 

Another issue we saw is that the first files per topic contain only one record, reason being lastRotate is zero and now - lastRotate >= rotateIntervalMs is always (or most likely) true.

I sent a PR to fix these issues.

Please have a look if it makes sense.

wjptak:
We have the same issue, using the latest Confluent Kafka. Thanks for the PR!!

cotedm:
This was resolved by #100

PR #100, opened 2016-08-12
flush the last partial file when incoming stream is paused, and not g…

skyahead:
…enerating first small files.

https://github.com/confluentinc/kafka-connect-hdfs/issues/79

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -73,6 +73,8 @@ public class TopicPartitionWriter {
   private int flushSize;
   private long rotateIntervalMs;
   private long lastRotate;
+  private boolean isFirst;
+  private boolean flushPartial;
   private long rotateScheduleIntervalMs;
   private long nextScheduledRotate;
   private RecordWriterProvider writerProvider;
@@ -176,6 +178,9 @@ public class TopicPartitionWriter {
     if(rotateScheduleIntervalMs > 0) {
       timeZone = DateTimeZone.forID(connectorConfig.getString(HdfsSinkConnectorConfig.TIMEZONE_CONFIG));
     }
+
+    setIsFirst(true);
+    setFlushPartial(true);
   }
 
   private enum State {
@@ -323,6 +328,22 @@ public class TopicPartitionWriter {
       }
     }
     if (buffer.isEmpty()) {
+      // committing files after waiting for rotateIntervalMs time but less than flush.size records available
+      if (flushPartial && recordCounter > 0 && shouldRotate(now)) {
+        log.info("committing files after waiting for rotateIntervalMs time but less than flush.size records available.");
+        updateRotationTimers();
+
+        try {
+          closeTempFile();
+          appendToWAL();
+          commitFile();
+        } catch (IOException e) {
+          log.error("Exception on topic partition {}: ", tp, e);
+          failureTime = System.currentTimeMillis();
+          setRetryTimeout(timeoutMs);
+        }
+      }
+
       resume();
       state = State.WRITE_STARTED;
     }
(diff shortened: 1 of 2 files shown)


skyahead:
[clabot:check]

skyahead:
[clabot:check]

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -398,10 +419,24 @@ private void setState(State state) {
     this.state = state;
   }
 
+  public void setIsFirst(boolean isFirst) {
+    this.isFirst = isFirst;
+  }
+
+  public void setFlushPartial(boolean flushPartial) {
```
ewencp: It seems that `flushPartial` is only set to `true` in practice and that the `setFlushPartial` call only exists to support the tests. Why have the `flushPartial` flag at all if it only takes on one value? It looks like the same is also true of `isFirst`.

Normally we'd just test the code as a blackbox, feeding it some input and validating that the expected output was generated. Is there a reason that wouldn't work here? It seems like you just want partial data to be written upon flush, which seems like it should be testable without manually exposing some internal flags.

skyahead: @ewencp Sorry for late replying and thanks for the review.

I removed the flushPartial flag and the non-blackbox test case.

Regarding isFirst, it is kept with the following reason. 

In the shouldRotate method, there is a periodicRotation check on the time difference between 'now' and 'lastRotate' like so: now - lastRotate >= rotateIntervalMs. The 'lastRotate' is always zero because it is first used in shouldRorate() and will only be updated later on in updateRotationTimers(). 

Without the isFirst flag, the very first batch always contains only one record.

ewencp:
ok to test

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -398,10 +417,20 @@ private void setState(State state) {
     this.state = state;
   }
 
+  private void setIsFirst(boolean isFirst) {
+    this.isFirst = isFirst;
```
ewencp: Not really critical, but this method doesn't seem necessary since it's just used internally to set a single variable.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -398,10 +417,20 @@ private void setState(State state) {
...
+  }
+
   private boolean shouldRotate(long now) {
-    boolean periodicRotation = rotateIntervalMs > 0 && now - lastRotate >= rotateIntervalMs;
+    boolean periodicRotation = rotateIntervalMs > 0 && (!isFirst && now - lastRotate >= rotateIntervalMs);
     boolean scheduledRotation = rotateScheduleIntervalMs > 0 && now >= nextScheduledRotate;
     boolean messageSizeRotation = recordCounter >= flushSize;
+
+    if (isFirst) {
```
ewencp: Seems like we could also avoid the whole `isFirst` business if we just initialize `lastRotate` to be the current time in the constructor.

ewencp:
@skyahead Since this LGTM but needed a bit more cleanup and retargeting to 3.1.x (which will merge into master), I made the minor updates in https://github.com/confluentinc/kafka-connect-hdfs/pull/134 with your commits preserved. Thanks for the contribution!

skyahead:
@ewencp Thanks very much for your help and terrible sorry for my slowness. Got quite snowed under by work (on other hdfs connector issues :-)). 

A reason that I was using the 'isFirst' flag is that sometime shouldRotate() can be true even if no records appeared. 

For example, if the recover() method (that talks to HDFS) takes longer than rotateIntervalMs time, then shouldRotate() will be true and am empty file will be generated. The unit test can not verify this situation but I saw this happening in our test cluster. Will try to confirm as soon as I can.

ewencp:
@skyahead Haha, no problem, I understand the feeling :)

Re: empty files, I think we should probably just handle empty files an additional case in `shouldRotate` then, i.e. actually track the amount of data written so far. This works more generally and is a bit less messy.

skyahead:
@ewencp Tracking the amount of actual data written is a great idea!

Issue #77, opened 2016-06-15
readOffset is slow/expensive for large, highly partitioned tables


jingw:
Our current setup has a `Partitioner` that routes data into `{topic}/day={...}/hour={...}/{actual data files}`. In some load testing with a 500 partition topic, we managed to DDoS HDFS's NameNode when our consumers were all recovering and calling `readOffset`. This function recursively scans the contents of the topic for each partition, performing a total of O(time \* partitions^2) work.

What do you feel would be a good solution to this?
1. Include offsets in WAL recovery: Have `wal.apply()` return the max offset, and then get rid of `resetOffsets`. The WAL interface would have to be less generic and would have to know about the file name format. If the WAL is empty then we need something else. Maybe part of WAL recovery can be a special "current offset" file.
2. Work around: Make Kafka partition the first directory in the `Partitioner` and do some magic special casing in `readOffset` and/or `fileStatusWithMaxOffset`. We could probably also assume that the latest offset isn't in an old time partition. If we do both of these tricks then total recovery cost becomes O(partitions \* constant for how far back we scan).

I'm happy to contribute a PR for this (though I imagine the work around in option 2 wouldn't be easily publishable).

Thanks!

Ishiihara:
@jingw Thanks for reporting this. We definitely want to improve this behavior. Do you want to elaborate a bit more of your both options? We need a good design here as so that the scalability of the connector can be improved.

jingw:
Thanks for the reply.

For (1), here's a patch I've been playing with: https://gist.github.com/jingw/5f427f09037c26a21978e299415e7d39 (not complete, not successfully tested)
The idea is that the write ahead log knows about the offsets, so we should be able to use that instead of recursively scanning HDFS.

(2) is a quick hack to reduce the scale of the problem. If I make my partitioning format partition/day/hour instead of day/hour, then each partition only needs to scan the day/hour for its own partition. Furthermore, if we're willing to assume some amount of ordering (e.g. data isn't out-of-order by more than a day), then we can assume the latest offset is in one of the day/hour directories. This approach would probably work for me, but I'm guessing it wouldn't be suitable for inclusion in your repo.

dosvath:
Addressed by https://github.com/confluentinc/kafka-connect-hdfs/pull/556

PR #107, opened 2016-08-25
Use hsync instead of flush

jingw:
I'm not a HDFS expert, but it looks like flush isn't implemented by DFSOutputStream, HdfsDataOutputStream, or FSDataOutputStream. The only methods declared by the Syncable interface are sync (deprecated), hflush, and hsync.

This should avoid some of the cases where the WAL is empty.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
+++ b/src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
@@ -315,7 +315,7 @@ public class WALFile {
         throws IOException {
       out.write(VERSION);                    // write the version
       out.write(sync);                       // write the sync bytes
-      out.flush();                           // flush header
+      out.hsync();                           // flush header
     }
 
 
@@ -328,7 +328,7 @@ public class WALFile {
         if (ownOutputStream) {
           out.close();
         } else {
-          out.flush();
+          out.hsync();
         }
         out = null;
       }


Ishiihara:
ok to test

ewencp:
@jingw I'm not sure I understand the issue. The `flush()` method isn't implemented directly by those classes, but they have implementations in parent classes as far as I can tell. For example, `DFSOutputStream` extends `FSOutputSummer` which implements `flush()`. Presumably they all *must* implement it since they implement `java.io.OutputStream`, which includes `flush()` in its interface.

If there are places where we need an `hflush` instead of `flush` (which seems to guarantee stronger semantics wrt HDFS readers, whereas `flush` I think only guarantees local buffers to be empty), or where we are using `flush`/`hflush` when we actually need `hsync` (to guarantee the write has fully been written to underlying storage), that could make sense but I'd like to avoid adding more expensive calls unless they are necessary.

jingw:
Got it, thanks. I was confused by why we were calling flush when it didn't seem to offer any guarantees on data visibility or persistence. This was mostly relevant to me, since I had a tweak that required getting the lease earlier in FSWAL.apply, before the attempt to play the WAL. At that point acquireLease would have written the header and flushed, but the reader wouldn't see any of it and would complain about an empty file. For the stock code here, this diff is supposed to prevent the WAL file from being empty if we die before the first hsync, which then causes all subsequent WAL reading to crash.

I'll just keep this tweak to myself for now (maybe bundling it with a bigger diff for #77).

Issue #98, opened 2016-08-02
TimeBasedPartitioner according to specified field


kerkero:
Please add the ability to have time based partitioning by specific field denoting the event time (instead of ingestion time)

xianzhen:
I used fieldPartioner to handle it. But i think your idea is better!

doriwal:
Agree that timeBased  should have the option to use event time , for example event that send from mobile SDK at 8:58 might be pushed to kafka at 9:01 , using the ingestion time is not accurate 

I also added pull request in order to support several fieldsPartitioner as in my use case I need both event time and event type in order to partition the events in parquet

ewencp:
@xianzhen `FieldPartitioner` only works for exact values, which means you can't use a timestamp field directly, you need it to be simplified to, e.g., minute, hour, or day, depending on the granularity you want.

@kerkero I think this would definitely be valuable. In particular, Kafka added timestamps natively in its format (but after this connector was originally created) so we should at least support those. Additionally, since Connect supports timestamp logical types, we should definitely have support for those fields. I think this should amount to a small additional config to specify the field and minor modifications to `TimeBasedPartitioner`.

doriwal:
Agree that FieldPartitioner cant use timestamp , in my use case each event that i am sending to kafka contains also the following fields : day like 2016-08-23 , time like 10:00 
this fields are calculated from the event timestamp (I am adding day/time fields to the event that i am sending to kafka)

codenamelxl:
A little head-up, it seems that kafka-connect-hdfs has been change to use [kafka-connect-storage-common](https://github.com/confluentinc/kafka-connect-storage-common) which does have a very nice way to provide a timestamp field ( it can be seen [here](http://docs.confluent.io/current/connect/connect-storage-cloud/kafka-connect-s3/docs/configuration_options.html) in kafka-connect-s3 which also use kafka-connec-storage-common. We now have to wait to this change come into release :)

kkonstantine:
Indeed, timestamp-based partitioners where timestamp is stored in the Kafka record or a record field have become available under `kafka-connect-storage-common`. 

Two recent PRs have added support of such partitioners in the HDFS connector (https://github.com/confluentinc/kafka-connect-hdfs/pull/245 and https://github.com/confluentinc/kafka-connect-hdfs/pull/246). 

[...]

PR #109, opened 2016-08-27
support specifying time field in TimeBasedPartitioner

simshi:
Try to solve #98

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -190,6 +190,12 @@ public class HdfsSinkConnectorConfig extends AbstractConfig {
   public static final String PARTITION_FIELD_NAME_DEFAULT = "";
   public static final String PARTITION_FIELD_NAME_DISPLAY = "Partition Field Name";
 
+  public static final String PARTITION_TIME_FIELD_NAME_CONFIG = "partition.timefield.name";
+  private static final String PARTITION_TIME_FIELD_NAME_DOC =
+          "The timestamp field name of SinkRecord when TimeBasedPartitioner is used, "
+          + "current time will be used on empty";
+  public static final String PARTITION_TIME_FIELD_NAME_DEFAULT = "";
+
   public static final String PARTITION_DURATION_MS_CONFIG = "partition.duration.ms";
   private static final String PARTITION_DURATION_MS_DOC =
       "The duration of a partition milliseconds used by ``TimeBasedPartitioner``. "
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -190,11 +190,10 @@ public class HdfsSinkConnectorConfig extends AbstractConfig {
   public static final String PARTITION_FIELD_NAME_DEFAULT = "";
   public static final String PARTITION_FIELD_NAME_DISPLAY = "Partition Field Name";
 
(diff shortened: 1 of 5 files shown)


simshi:
Hi robot, 
[clabot:check]

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -190,6 +190,12 @@
   public static final String PARTITION_FIELD_NAME_DEFAULT = "";
   public static final String PARTITION_FIELD_NAME_DISPLAY = "Partition Field Name";
 
+  public static final String PARTITION_TIME_FIELD_NAME_CONFIG = "partition.timefield.name";
```
ewencp: Normally we'd separate each word in the config name with a `.` (as you separate each in the variable name with a `_`), so `"partition.time.field.name"` would probably be better here.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -190,6 +190,12 @@
...
+  private static final String PARTITION_TIME_FIELD_NAME_DOC =
+          "The timestamp field name of SinkRecord when TimeBasedPartitioner is used, "
+          + "current time will be used on empty";
+  public static final String PARTITION_TIME_FIELD_NAME_DEFAULT = "";
```
ewencp: Can we leave it without a default instead? This is more natural and allows for better error checking as you can tell the user if they specify an empty (and therefore invalid) setting (which they may do, e.g., accidentally and need to clean up their config).

src/main/java/io/confluent/connect/hdfs/partitioner/TimeBasedPartitioner.java
```diff
@@ -86,14 +88,23 @@ public void configure(Map<String, Object> config) {
...
   }
 
   @Override
   public String encodePartition(SinkRecord sinkRecord) {
-    long timestamp = System.currentTimeMillis();
+    long timestamp;
+    if (timeFieldName.equals("")) {
+      timestamp = System.currentTimeMillis();
+    } else {
+      Struct struct = (Struct) sinkRecord.value();
```
ewencp: Probably want to validate the type before casting so you can throw the appropriate `ConnectException` if it isn't a `Struct`. Same deal for checking whether the field exists and if it's type is correct.

src/main/java/io/confluent/connect/hdfs/partitioner/TimeBasedPartitioner.java
```diff
@@ -86,14 +88,23 @@ public void configure(Map<String, Object> config) {
...
+      timestamp = (long) struct.get(timeFieldName);
```
ewencp: There are also logical types (e.g. `Timestamp`) that you might want to support here.

simshi:
Thank you, agreed on most of your comments except I don't know how to handle logical types like `Timestamp`. Support this feature in another PR? :smile:

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
simshi: My first attempt on eliminating default value failed, looked into `ConfigDef.parse()`, can't find a way to express "this config key is required only when partitioner.class is `TimeBasedPartitioner`". Suggestions are welcomed.

jocelyndrean:
Hey ! Thanks @simshi for your PR.  No news on it since Aug 30 ? No merge ? :/

simshi:
I'm not working on this, discarded, please roll your own version 😄

PR #113, opened 2016-09-07
Allow nameservice config for HA hdfs cluster implementations 

rnpridgeon:


rnpridgeon:
@ewencp mind taking a look

gwenshap:
LGTM. Since some configuration changes are required, maybe add a doc section that explains how to set the connector up for HA?

rnpridgeon:
Yeah we will definitely need to add something in the docs

rnpridgeon:
Looks like I was wrong this actually does work as is when you provide connect with both the core-site.xml and hdfs-site.xml in the hadoop properties file. No other fix is needed, when I reproduced the issue I wasn't picking up core-site.xml so I'm closing this since is not actually an issue. I'll go fix the docs instead

PR #117, opened 2016-09-16
CC-310: Always explicitly request assigned topic partitions to seek to an offset to ensure offsets committed to Kafka are not used when no files have been committed to HDFS yet.

ewencp:


ewencp:
@Ishiihara @confluentinc/connect @gwenshap System test failure and possible data loss issue. Here's a summary of the issue from the bug:

I've tracked down the source of this issue. All the errors are for the first range of offsets. The reason is that we can have cases where a rebalance starts, the we stop the task, it discards its current data but isn't rewinding offsets (which it shouldn't really need to do since anything recovering offsets should be looking in HDFS for committed offsets), but then the Kafka offsets are committed by the WorkerSinkTask during shutdown. flush() is called during that process, but because the partitions have already been revoked and data discarded there isn't even anything we could track anymore. When the next task starts up, it is looking for committed data and since it doesn't find any, the current behavior is for it not to do anything and just start using the offset provided by the consumer. However, in this case that is incorrect because we actually want to start from the earliest possible offset even if something was committed to the Kafka offsets.

This actually wouldn't be a problem with https://issues.apache.org/jira/browse/KAFKA-3462 but because empty list of offsets does not force an offset reset, during this early phase before the first file has been rotated, we are at risk of skipping that data. Until we get KAFKA-3462 in, the right behavior is to detect the empty case, reset the offset to 0, and that will force it to either start from the true beginning of the topic partition or be out of range and trigger a reset. In other words, we should make sure that code path always resets the offsets so the connector always controls what offsets we start reading from.

shikhar:
LGTM

PR #126, opened 2016-09-24
Fix TimeBasedPartitioner config validation

cotedm:
This PR is to fix the TimeBasedPartitioner validation.  Today it returns visible=false for ``partition.duration.ms`` if you call config/validate with the TimeBasedPartitioner set.  @shikhar can you review?


shikhar:
LGTM

PR #128, opened 2016-09-28
minor: remove end offset and clear temp file hash after committing files

terrencehan:
I observed end offsets are keeping growing via log[1]. IMHO it's should be cleaned after committing finished.

[1] https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L294


savulchik:
@terrencehan I observe them growing, too.

Issue #141, opened 2016-10-21
Exceptions when network is broken


skyahead:
I have a hdfs connector running on my machine, which reads from a Kafka Cluster and then writes into a HDFS cluster. When I disable my machine's network connection, and then enable it after a while, various exceptions throw. 

The problem is the hdfs connector will keep trying to acquire lease on WAL files **for ever, but never get it**.

The exceptions that the hdfs connector throws are:

> org.apache.kafka.connect.errors.ConnectException: Cannot acquire lease after timeout, will retry.
>         at io.confluent.connect.hdfs.wal.FSWAL.acquireLease(FSWAL.java:95) ~[main/:na]
>         at io.confluent.connect.hdfs.wal.FSWAL.apply(FSWAL.java:105) ~[main/:na]
>         at io.confluent.connect.hdfs.TopicPartitionWriter.applyWAL(TopicPartitionWriter.java:484) [main/:na]
>         at io.confluent.connect.hdfs.TopicPartitionWriter.recover(TopicPartitionWriter.java:212) [main/:na]
>         at io.confluent.connect.hdfs.TopicPartitionWriter.write(TopicPartitionWriter.java:256) [main/:na]
>         at io.confluent.connect.hdfs.DataWriter.write(DataWriter.java:234) [main/:na]
>         at io.confluent.connect.hdfs.HdfsSinkTask.put(HdfsSinkTask.java:103) [main/:na]
>         at org.apache.kafka.connect.runtime.WorkerSinkTask.deliverMessages(WorkerSinkTask.java:384) [main/:na]
>         at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:240) [main/:na]
>         at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:172) [main/:na]
>         at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:143) [main/:na]
>         at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:140) [connect-runtime-0.10.0.1-cp1.jar:na]
>         at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:175) [connect-runtime-0.10.0.1-cp1.jar:na]
>         at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [na:1.8.0_101]
[...]

PR #142, opened 2016-10-21
not trying to hold leases on WAL files if we are holding them already.

skyahead:
This PR is for: https://github.com/confluentinc/kafka-connect-hdfs/issues/141


lakeofsand:
@skyahead We meet the same question recently.
But,with log's traceback,we found our question is trigger by a close failed:

ERROR Error closing hdfs://192.168.101.55:8020/logs/******/1/log.
org.apache.kafka.connect.errors.ConnectException: Error closing hdfs://192.168.101.55:8020/logs/*****/1/log
at io.confluent.connect.hdfs.wal.FSWAL.close(FSWAL.java:156)
at io.confluent.connect.hdfs.TopicPartitionWriter.close(TopicPartitionWriter.java:325)
at io.confluent.connect.hdfs.DataWriter.onPartitionsRevoked(DataWriter.java:318)
at io.confluent.connect.hdfs.HdfsSinkTask.onPartitionsRevoked(HdfsSinkTask.java:108)
...
Caused by: java.io.InterruptedIOException: Interrupted while waiting for data to be acknowledged by pipeline
        at org.apache.hadoop.hdfs.DFSOutputStream.waitForAckedSeqno(DFSOutputStream.java:2151)
        at org.apache.hadoop.hdfs.DFSOutputStream.flushInternal(DFSOutputStream.java:2132)
        at org.apache.hadoop.hdfs.DFSOutputStream.close(DFSOutputStream.java:2230)
        at org.apache.hadoop.fs.FSDataOutputStream$PositionCache.close(FSDataOutputStream.java:72)
        at org.apache.hadoop.fs.FSDataOutputStream.close(FSDataOutputStream.java:106)
        at io.confluent.connect.hdfs.wal.WALFile$Writer.close(WALFile.java:329)
        at io.confluent.connect.hdfs.wal.FSWAL.close(FSWAL.java:150)

skyahead:
@lakeofsand I do not see exactly the same exceptions, but I do see lots of similar ones :-)

But I think after all those 'Error closing' exceptions, there is a 'topicPartitionWriters.remove(tp)' call in DataWriter.java's close() method, which will remove this writer and a new one will be recreated for the current partition. I.e., the connect code should survive these exceptions. Do you see your coding keeps generating these errors for ever?

But, It seems you are using a version that is older than this commit: [https://github.com/confluentinc/kafka-connect-hdfs/commit/b2b1c619120764e52808192717d849418659bcca#diff-d4f63c72e615f6185c4d472918ba1e95](https://github.com/confluentinc/kafka-connect-hdfs/commit/b2b1c619120764e52808192717d849418659bcca#diff-d4f63c72e615f6185c4d472918ba1e95), and the close() method is called 
onPartitionsRevoked() in your version.

lakeofsand:
Our issue seems like because of some bug in hdfs client(): @skyahead 

org.apache.hadoop.hdfs.DFSOutputStream.java:
@Override
  public void close() throws IOException {
    synchronized (this) {
      TraceScope scope = dfsClient.getPathTraceScope("DFSOutputStream#close",
          src);
      try {
        closeImpl();    ==> **the  previous "FSWAL.close" exception is throw here,dfsClient.endFileLease is missed.**
      } finally {
        scope.close();
      }
    }
    dfsClient.endFileLease(fileId);
  }

skyahead:
@lakeofsand When your errors happen, do you see anything wrong in your HDFS namenode log file?

lakeofsand:
@skyahead 
The "close exception" lead to a lease-renewer continue exist in connect process，HDFS namenode would not konw about it. File lease is continuely renewed by connect periodically.

Neither the process itself nor other process‘s  opertions to this file will be failed  for the lease is still owned by the previous process.

"
Failed to APPEND_FILE /logs/beaver_http_response/1/log for DFSClient_NONMAPREDUCE_450119871_31 on 192.168.101.101 because this file lease is currently owned by DFSClient_NONMAPREDUCE_1501121577_33 on 192.168.101.102
"

cotedm:
@skyahead it's definitely dangerous to parse an exception message (these can change without warning if we say upgrade the HDFS client dependency). However, the symptom you describe in issue #141 looks like the writer isn't set to null which happens here
https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java#L154

It seems possible that the writer didn't close properly in your scenario because you cut the network. In which case we thrown an exception and never null out the writer, so we don't ever create a new writer (thus obtaining a new lease). Maybe a better approach is to null out the writer and reader in a `finally` block in the close() method presented above.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -96,6 +97,11 @@ public void acquireLease() throws ConnectException {
     }
   }
 
+  private void reset() {
```
cotedm: Instead of adding a new method, could we just do this in the `close()` method in a finally block so that we attempt to close the reader/writer gracefully first before nulling them out? Then replace calls to `reset()` with `close()` in this file instead?

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -194,40 +197,52 @@ public static Option blockSize(long value) {
...
-          this.appendMode = true;
         } else {
-          out = fs.create(p, true, bufferSize, replication, blockSize);
+          out = streamOption.getValue();
         }
-      } else {
-        out = streamOption.getValue();
+
+        init(conf, out, ownStream);
+      } catch (RemoteException re) {
```
cotedm: Can you explain this part a bit more? I'm not sure why we need to look for this exception here but I might just be missing it.

src/test/java/io/confluent/connect/hdfs/wal/WALTest.java
```diff
@@ -73,6 +73,7 @@ public void run() {
     });
     thread.start();
 
+    Thread.sleep(3001);
```
cotedm: this seems unrelated?

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -508,25 +524,38 @@ public Reader(Configuration conf, Option... opts) throws IOException {
...
+            : lenOpt.getValue();
+          file = openFile(fs, filename, bufSize, len);
+        } else {
+          len = null == lenOpt ? Long.MAX_VALUE : lenOpt.getValue();
+          file = streamOpt.getValue();
+        }
+        long start = startOpt == null ? 0 : startOpt.getValue();
+        // really set up
+        initialize(filename, file, start, len, conf, headerOnly != null);
+      } catch (RemoteException re) {
```
cotedm: Same as the other RemoteException, can you explain why we need to catch this?

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -60,12 +61,16 @@ public void append(String tempFile, String committedFile) throws ConnectExceptio
       writer.append(key, value);
       writer.hsync();
     } catch (IOException e) {
+      close();
```
ewencp: Hmm, so just for the sake of debuggability, if we're doing work in the exception handler that could itself throw exceptions, I think we might want to at least log that something went wrong. Because with the current code, if `close()` throws an exception, then all the info about the original exception is lost.

This is true for a couple of the other similar changes below. The only alternative I could think of is allowing `close()` to take an extra parameter that's the original cause of calling close (or null if there isn't one) and attaching that as the cause if it fails, but then we also lose info about the other exception. I think logging this one proactively is probably fine since these should all be exceptional conditions.

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
ewencp: Yeah, I can't remember the details but I think the issue is which jar the class is defined in or something like that. So we had to go down the hacky route of checking the class name.

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -194,40 +197,52 @@ public static Option blockSize(long value) {
...
         } else {
-          out = fs.create(p, true, bufferSize, replication, blockSize);
+          out = streamOption.getValue();
         }
-      } else {
-        out = streamOption.getValue();
+
+        init(conf, out, ownStream);
+      } catch (RemoteException re) {
+        log.error("Failed creating a WAL Writer: " + re.getMessage());
```
ewencp: Can we do
```
log.error("Failed creating a WAL Writer: ", re);
```
instead so the log will include stack trace info?

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -194,40 +197,52 @@ public static Option blockSize(long value) {
...
+        if (re.getClassName().equals(leaseException)) {
+          if (fs != null) {
+            fs.close();
```
ewencp: Does this need protection from a possible `IOException`?

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
ewencp: In fact, is the log even going to be useful since it gets logged again by `FSWAL` which calls this constructor? I think this might be the case in the other case below as well, though I haven't extensively checked all callers. I guess worst case we're just logging a bit more information, so perhaps being conservative about logging the details isn't a bad idea.

src/main/java/io/confluent/connect/hdfs/wal/WAL.java
```diff
@@ -27,4 +27,5 @@
   void truncate() throws ConnectException;
   void close() throws ConnectException;
   String getLogFile();
+  long getSleepIntervalMs();
```
ewencp: Hmm, a bit of a nit, but since this is only used for testing I'm not sure about making it public for testing. In fact, we separately have some work going on to refactor some of this code to make it more reusable and this seems like an odd addition to the interface given that it's really just public for testing. /cc @kkonstantine 

I'm wondering if just casting to the more specific FSWAL class in the test would be a better solution (and label the FSWAL public API as public for testing)? Keeps the interface clean but allows the test to access the info that it needs.

skyahead: great idea

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
skyahead: Yes, I was hoping to log a bit more information so that it is easier to read the debug log which is VERY long. 

So re.getMessage() is the short version that I like :-) If we log the stack trace, the same info will be logged again all the way up at https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L324.

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
skyahead: Nope. I think IOException will be caught here: https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L324.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
skyahead: Please have a look now.

cotedm:
I don't have any other comments, what do you think @ewencp ?

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -60,12 +61,17 @@ public void append(String tempFile, String committedFile) throws ConnectExceptio
       writer.append(key, value);
       writer.hsync();
     } catch (IOException e) {
+      log.error("Error appending WAL file: {}, {}", logFile, e);
+      close();
       throw new ConnectException(e);
     }
   }
 
+  public long getSleepIntervalMs() {
```
kkonstantine: nit: Could have "package private" scope instead and a non-javadoc comment above to show it's destined for testing.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -143,6 +151,8 @@ public void truncate() throws ConnectException {
       close();
```
kkonstantine: Redundant call, after `close` has been added in `finally` block.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -159,6 +169,9 @@ public void close() throws ConnectException {
       }
     } catch (IOException e) {
       throw new ConnectException("Error closing " + logFile, e);
+    } finally {
+      writer = null;
```
kkonstantine: I can't point to the actual lines, but nullifying reader and writer above, is redundant now that it's done within the `finally` block.

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -56,6 +57,8 @@
   private static final int SYNC_HASH_SIZE = 16;   // number of bytes in hash
   private static final int SYNC_SIZE = 4 + SYNC_HASH_SIZE; // escape + hash
 
+  private static final String leaseException = "org.apache.hadoop.hdfs.protocol.AlreadyBeingCreatedException";
```
kkonstantine: nit: It'd be nice for variable naming to be consistent with our code style for `static final` member fields. 
Thus, it should be something like: `LEASE_EXCEPTION_CLASS_NAME`
(I added more to the name to show it's a class name string and not to actual class. Your call). 

I know we don't run checkstyle currently, but we will soon.

src/test/java/io/confluent/connect/hdfs/wal/WALTest.java
```diff
@@ -43,7 +43,7 @@ public void testWALMultiClient() throws Exception {
     Storage storage = StorageFactory.createStorage(storageClass, conf, url);
 
     final WAL wal1 = storage.wal(topicsDir, TOPIC_PARTITION);
-    final WAL wal2 = storage.wal(topicsDir, TOPIC_PARTITION);
+    final FSWAL wal2 = (FSWAL)storage.wal(topicsDir, TOPIC_PARTITION);
```
kkonstantine: Single white space needed between casting and `storage`

src/test/java/io/confluent/connect/hdfs/wal/WALTest.java
```diff
@@ -62,8 +62,8 @@ public void testWALMultiClient() throws Exception {
       @Override
       public void run() {
         try {
-          // holding the lease for awhile
-          Thread.sleep(3000);
+          // holding the lease for time that is less than wal2's retry interval, which is 1000 ms.
+          Thread.sleep(wal2.getSleepIntervalMs()-100);
```
kkonstantine: White space needed around `-`
(Again checkstyle coming soon :) )

src/test/java/io/confluent/connect/hdfs/wal/WALTest.java
```diff
@@ -73,6 +73,8 @@ public void run() {
     });
     thread.start();
 
+    // AcquireLease will try to acquire the same lease that wal1 is holding and fail
```
kkonstantine: nit: I'd use the actual method name (`acquireLease`) if I wanted to refer to it, instead of the concept ("Acquiring the lease")

skyahead:
@kkonstantine Thanks for the review! how about now?

kkonstantine:
Great, thanks. I see you also pushed the declaration of initial and max intervals as constants outside the method, I forgot to mention that. 

Ok by me, I'd wait for final comments from @ewencp.

skyahead:
@kkonstantine thanks for replying so promptly :-)

skyahead:
@ewencp Can you have a final look at this PR please?

Perdjesk:
@ewencp @kkonstantine 
I hope everyone is doing well. Is there anything that can help to go forward with this PR.

invis87:
Good day for everyone!
I have the same issue in my cluster and need to execute `hdfs debug recoverLease -path ${path}` every day...
Looks like you fixed this issue for a long time ago, so why it is still not merged?

ewencp:
ok to test

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -35,15 +35,13 @@
...
-  private static final String leaseException =
-      "org.apache.hadoop.hdfs.protocol.AlreadyBeingCreatedException";
-  private static final long MAX_SLEEP_INTERVAL_MS = 16000L;
 
   private WALFile.Writer writer = null;
   private WALFile.Reader reader = null;
   private String logFile = null;
   private HdfsSinkConnectorConfig conf = null;
   private HdfsStorage storage = null;
+  private long sleepIntervalMs = WALConstants.INITIAL_SLEEP_INTERVAL_MS;
```
kkonstantine: Why is this one upgraded from local variable to member field variable? The behavior definitely changes on repeated calls on the same object. Notice that this variable is mutated iteratively as part of the `while` loop in `acquiredLease`. Even if it happens and there's only one call of `acquireLease` per `FSWAL` object I think it's a better practice to keep it local if that's what we intend to do here.

src/main/java/io/confluent/connect/hdfs/wal/WALConstants.java
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright 2015 Confluent Inc.
```
kkonstantine: nit: If you apply any changes based on the comment above, you may change this one too to be `2017` instead. Otherwise, never mind.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
skyahead: Good catch!

PR #154, opened 2016-11-23
Do not get stuck on compacted logs

rossoft:
Connect gets stuck on topics with compacted logs. If we try to process a partition that have been compacted, TopicPartitionWriter gets stuck and keeps discarding records forever, as it will never find the expected offset.

As the expected offset may be deleted in the partition, it should just expect to see any record with an offset equal or greater than the expected one.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -80,7 +80,7 @@ public class TopicPartitionWriter {
   private AvroData avroData;
   private Set<String> appended;
   private long offset;
-  private boolean sawInvalidOffset;
+  private boolean sawStaleOffset;
   private Map<String, Long> startOffsets;
   private Map<String, Long> offsets;
   private long timeoutMs;
@@ -156,7 +156,7 @@ public class TopicPartitionWriter {
     state = State.RECOVERY_STARTED;
     failureTime = -1L;
     offset = -1L;
-    sawInvalidOffset = false;
+    sawStaleOffset = false;
     extension = writerProvider.getExtension();
     zeroPadOffsetFormat
         = "%0" +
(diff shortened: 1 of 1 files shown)


ewencp:
ok to test

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -511,23 +511,23 @@ private void writeRecord(SinkRecord record) throws IOException {
     long expectedOffset = offset + recordCounter;
     if (offset == -1) {
       offset = record.kafkaOffset();
-    } else if (record.kafkaOffset() != expectedOffset) {
+    } else if (record.kafkaOffset() < expectedOffset) {
```
ewencp: Hmm, 2 things. First, I think if we change this then we want `expectedOffset` to be computed differently. Right now it just adds the record count, but that wouldn't make sense anymore.

Secondly, maybe we can just remove this entirely? Or just move the check behind a config flag maybe so there's a way to turn it off if you don't need it? The underlying issue that required this check is fixed in 0.10.1.0, although removing this entirely would put folks trying to run a newer version of the connector in an older cluster at risk.

rossoft: Well I think that  less-than condition will work for new version where this check is not needed at all, and is backwards compatible as well. It is essentially how the original condition should have been since it was expecting "rewinds" back in the partition but not "fast forwards" with holes.
Another option as you say is to just protect this under a configuration setting. That works for me as well. So you decide, if you want, I can change the PR to leave the original condition under a config flag.

ewencp: I think the logic doesn't quite work based on how `recordCounter` is updated though. You're right that on the first message this works fine. But let's say you start a new `TopicPartitionWriter` at offset 100, but actually messages 100-199 have been compacted. So you see offset 200. This check will work for that first message, but then the next iteration you'll have incremented `recordCounter` by 1 and `expectedOffset will be 101. While it will allow valid messages through, there's a gap of messages from 102-200 that wouldn't be properly rejected. So I think we'd need to get rid of recordCounter and just update a variable with the last seen offset to validate against.

re: the config setting, I think we will need it if we can't guarantee bad data won't be allowed through accidentally. If we can figure out a version that works and handles that bug, I'm all for doing it without the config setting.

rossoft: I have changed it back to the original logic + a config flag to disable it

PR #172, opened 2017-02-16
Support specifying time field in TimeBasedPartitioner. 

westerly:
Tried to solve #98. 

I have started from there https://github.com/confluentinc/kafka-connect-hdfs/pull/109.
I have also added support for time fields located in nested structs of a sink record.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -14,20 +14,6 @@
 
 package io.confluent.connect.hdfs;
 
-import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.FileStatus;
-import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.security.SecurityUtil;
-import org.apache.hadoop.security.UserGroupInformation;
-import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.config.ConfigException;
-import org.apache.kafka.connect.data.Schema;
-import org.apache.kafka.connect.errors.ConnectException;
-import org.apache.kafka.connect.sink.SinkRecord;
-import org.apache.kafka.connect.sink.SinkTaskContext;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.io.IOException;
 import java.net.InetAddress;
 import java.util.Collection;
@@ -44,6 +30,20 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.security.SecurityUtil;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.apache.kafka.connect.sink.SinkTaskContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import io.confluent.connect.avro.AvroData;
 import io.confluent.connect.hdfs.filter.CommittedFileFilter;
@@ -402,6 +402,7 @@ public class DataWriter {
   private Map<String, Object> copyConfig(HdfsSinkConnectorConfig config) {
     Map<String, Object> map = new HashMap<>();
     map.put(HdfsSinkConnectorConfig.PARTITION_FIELD_NAME_CONFIG, config.getString(HdfsSinkConnectorConfig.PARTITION_FIELD_NAME_CONFIG));
+    map.put(HdfsSinkConnectorConfig.PARTITION_TIME_FIELD_NAME_CONFIG, config.getString(HdfsSinkConnectorConfig.PARTITION_TIME_FIELD_NAME_CONFIG));
     map.put(HdfsSinkConnectorConfig.PARTITION_DURATION_MS_CONFIG, config.getLong(HdfsSinkConnectorConfig.PARTITION_DURATION_MS_CONFIG));
     map.put(HdfsSinkConnectorConfig.PATH_FORMAT_CONFIG, config.getString(HdfsSinkConnectorConfig.PATH_FORMAT_CONFIG));
     map.put(HdfsSinkConnectorConfig.LOCALE_CONFIG, config.getString(HdfsSinkConnectorConfig.LOCALE_CONFIG));
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -14,17 +14,17 @@
 
 package io.confluent.connect.hdfs;
 
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
 import org.apache.kafka.common.config.ConfigException;
-
-import java.util.Arrays;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
 
 import io.confluent.connect.hdfs.partitioner.DailyPartitioner;
 import io.confluent.connect.hdfs.partitioner.DefaultPartitioner;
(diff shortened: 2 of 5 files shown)


westerly:
Hi robot,
[clabot:check]

westerly:
Hi robot,
[clabot:check]

CBR09:
@westerly Thanks you for your PR, any update on it?

codenamelxl:
Hi, @westerly,
Thanks for the PR, it would be really nice to have this feature. Have you tested that this is working properly? I think i might just go ahead and compile this to use while waiting for this patch come into release.
@ewencp I see that you commented on the previous PR of this feature, will you have time to take a look at this and merge it anytime soon? Thank you very much.

gregroberti:
Is there a plan to merge this PR? I would like to use it 😄

codenamelxl:
A little head-up [https://github.com/confluentinc/kafka-connect-hdfs/issues/98#issuecomment-325543307](https://github.com/confluentinc/kafka-connect-hdfs/issues/98#issuecomment-325543307)

cricket007:
At first glance I think this will help me with #229 . I'm surprised Camus is considered deprecated, but this functionality wasn't carried over, or seems to be documented from what I can find without scanning the code.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -45,6 +31,20 @@
...
+import org.apache.hadoop.security.UserGroupInformation;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.apache.kafka.connect.sink.SinkTaskContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
```
rhauch: It'd be great to avoid all of these unnecessary changes.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -190,6 +190,13 @@
   public static final String PARTITION_FIELD_NAME_DEFAULT = "";
   public static final String PARTITION_FIELD_NAME_DISPLAY = "Partition Field Name";
 
+  public static final String PARTITION_TIME_FIELD_NAME_CONFIG = "partition.time.field.name";
+  private static final String PARTITION_TIME_FIELD_NAME_DOC =
+          "The timestamp field name of SinkRecord when TimeBasedPartitioner is used, "
+          + "current time will be used on empty";
```
rhauch: How about the following reworded description?

> The name of the field in the sink record that contains the timestamp as milliseconds past epoch to be used for partitioning when TimeBasedPartition used. When this is empty, the partitioning will use the current time the record is processed.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -314,19 +321,21 @@
...
   }
 
   private static class SchemaCompatibilityRecommender extends BooleanParentRecommender {
-    
+
     public SchemaCompatibilityRecommender() {
       super(HIVE_INTEGRATION_CONFIG);
     }
-      
+
```
rhauch: Again, it'd be great to avoid changes like these that are unrelated to the focus of the PR.

src/main/java/io/confluent/connect/hdfs/partitioner/TimeBasedPartitioner.java
```diff
@@ -14,29 +14,32 @@
...
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
```
rhauch: More unrelated changes.

src/main/java/io/confluent/connect/hdfs/partitioner/TimeBasedPartitioner.java
```diff
@@ -86,16 +89,53 @@ public void configure(Map<String, Object> config) {
     String hiveIntString = (String) config.get(HdfsSinkConnectorConfig.HIVE_INTEGRATION_CONFIG);
     boolean hiveIntegration = hiveIntString != null && hiveIntString.toLowerCase().equals("true");
 
+    timeFieldName = (String) config.get(HdfsSinkConnectorConfig.PARTITION_TIME_FIELD_NAME_CONFIG);
+
```
rhauch: How about checking this field name and setting it to null if the field name is empty or is blank when trimmed? That way the field name is either null or contains a name.

src/main/java/io/confluent/connect/hdfs/partitioner/TimeBasedPartitioner.java
```diff
@@ -86,16 +89,53 @@ public void configure(Map<String, Object> config) {
...
     init(partitionDurationMs, pathFormat, locale, timeZone, hiveIntegration);
   }
 
   @Override
   public String encodePartition(SinkRecord sinkRecord) {
-    long timestamp = System.currentTimeMillis();
-    DateTime bucket = new DateTime(getPartition(partitionDurationMs, timestamp, formatter.getZone()));
-    return bucket.toString(formatter);
+      long timestamp;
+      if (timeFieldName == null || timeFieldName.equals("")) {
```
rhauch: The `.equals("")` is not terribly efficient. By changing the `timeFieldName` to be null when not applicable (see previous comment), this if block becomes simpler and can just check whether the time field name is null.

src/main/java/io/confluent/connect/hdfs/partitioner/TimeBasedPartitioner.java
```diff
@@ -86,16 +89,53 @@ public void configure(Map<String, Object> config) {
...
+          timestamp = System.currentTimeMillis();
+      } else if (sinkRecord.value() instanceof Struct) {
+          Struct struct = (Struct) sinkRecord.value();
+          timestamp = getTimestampValue(struct);
+      } else {
+          throw new DataException("sinkRecord parameter is expected to be an instance of Struct");
```
rhauch: How about changing the message to be a bit more clear? Perhaps "The value of the sink record is expected to be a Struct type"?

src/main/java/io/confluent/connect/hdfs/partitioner/TimeBasedPartitioner.java
```diff
@@ -86,16 +89,53 @@ public void configure(Map<String, Object> config) {
...
+      }
+      DateTime bucket = new DateTime(getPartition(partitionDurationMs, timestamp, formatter.getZone()));
+      return bucket.toString(formatter);
+  }
+
+  private long getTimestampValue(Struct record) {
```
rhauch: Would be good to add unit tests for this method as well as `encodePartition(...)`. I know that the new unit tests you added will use these methods and therefore test them indirectly, but it's not easy with those kinds of indirect tests to know that the tests cover all the logic in the methods. That's much easier to do with unit tests that directly test the methods. You'd have to make the methods protected, but that's okay.

src/main/java/io/confluent/connect/hdfs/partitioner/TimeBasedPartitioner.java
```diff
@@ -86,16 +89,53 @@ public void configure(Map<String, Object> config) {
...
+      Struct tmpRecord = record;
+      Object tmpRecordObject = null;
+      while (i < fieldNames.length - 1) {
+          tmpRecordObject = getField(tmpRecord, fieldNames[i]);
+          tmpRecord = (Struct) tmpRecordObject;
+          i++;
+      }
+      tmpRecordObject = getField(tmpRecord, fieldNames[i]);
+      if (!(tmpRecordObject instanceof Long)) {
+          throw new DataException("The value of the time field must be a Long.");
```
rhauch: Would it be valid to check to see if the `tmpRecordObj` is a `Number`, and then call `Number.longValue()`? That way, this logic would work even if the data happens to be encoded as a long, integer, floating point number, or even BigInteger or BigDecimal. It is true that most of the time the value should be a long for semantic reasons, but being a bit more flexible would not seem to hurt.

cricket007:
I've pulled this code into the 3.2.x release code and tried it out. Works great!

I've also addressed all your comments, @rhauch . Would I need to push to `westerly:master`, or can I somehow push to this PR?

rhauch:
@cricket007 excellent work. How about creating a new PR from your `3.2.x`-based branch? We prefer to have PRs for the oldest branches to be changed, and then we merge them onto the newer branches like `3.3.x`, `3.4.x` and `master`.

cricket007:
@rhauch Done. Hopefully `3.2.2-post` branch is okay rather than `3.2.x`

kkonstantine:
This PR seems superseded by recent PRs here: https://github.com/confluentinc/kafka-connect-hdfs/pull/245 and here: https://github.com/confluentinc/kafka-connect-hdfs/pull/246

Should we close it?

rhauch:
@kkonstantine #245 and @246 were merged only on the `4.0.x` and will be included in the 4.0.0 release. However, since this PR were originally opened prior to the refactoring on storage-common for the `4.0.x` branch, it'd be possible to update this PR to (a) rebase it onto the `3.2.2` branch and (b) align it with the 4.0.0 configurations, and (c) then merged back for the next 3.2.x and 3.3.x releases. Meanwhile, the 4.0.0 release will contain the already-merged feature set from storage-common. WDYT?

kkonstantine:
Allow me to start by saying that I see this PR more like it's introducing a new feature rather than addressing a regression. Given that, I believe it shouldn't be backported. 

Then, by quickly checking the code, I see that the changes affect only `encodePartition` and having implemented time-based rotation on S3 Connector and HDFS connector > 4.0.x I can say that this might work under some cases but it's not a general solution. 

To give you an example, what happens when you reprocess a set of records? My understanding is that without changing the DataWriter to base its rotation decision on the partitioner instead of just the wall-clock, this will break exactly-once semantics in cases that wouldn't want to use wall-clock based partitioning. 

But also, by changing only `encodePartition` I don't think the granularity of `rotate.interval.ms` is taken in consideration in all cases. The above "happens" to work when your format matches the rotation interval, but won't work in other cases. 

Again, back to my first argument, I don't think this is a regression. The additional work that is needed to make timestamp based rotation doable has been put in the PRs I attached above, and is present in versions >= 4.x.x

rhauch:
Closing per @kkonstantine analysis. Note that this is already implemented in 4.0.0 (in the `4.0.x` and `master` branches) as noted above.

PR #179, opened 2017-03-13
CC-437: Update 3.2.0 changelog

ewencp:


ewencp:
cc @confluentinc/connect

docs/changelog.rst
```diff
@@ -3,6 +3,12 @@
 Changelog
 =========
 
+Version 3.2.0
+-------------
+
+* `PR-135 <https://github.com/confluentinc/kafka-connect-hdfs/pull/135>`_ - Fix typos
+* `PR-164 <https://github.com/confluentinc/kafka-connect-hdfs/pull/164>`_ - Issue 136 - Support topic with dots in hive.
+
```
ewencp: Yeah, the intent is to catch all the merged PRs for the release, and then I filter really pointless ones (e.g. when people are just merging version branches). I think this was merged without a merge commit, which loses the commit log pattern I have a script searching for. (We need a more robust way to generate these, but searching for the PR merge commits works well for most of our repos/PRs currently.)

kkonstantine: I see, that's why. Makes sense.

PR #188, opened 2017-04-26
CC-397: Refactoring on top of storage-common

kkonstantine:


src/main/java/io/confluent/connect/hdfs/hive/HiveMetaStore.java
```diff
@@ -1,12 +1,12 @@
 /**
  * Copyright 2015 Confluent Inc.
- *
+ * <p>
```
ewencp: I don't think this was intended (and may eventually fail rat checks).

src/main/java/io/confluent/connect/hdfs/hive/HiveSchemaConverter.java
```diff
@@ -14,81 +14,6 @@
...
-  public static TypeInfo convertMap(Schema schema) {
-    return TypeInfoFactory.getMapTypeInfo(
-        convert(schema.keySchema()), convert(schema.valueSchema()));
-  }
-
-  public static TypeInfo convertPrimitive(Schema schema) {
-    return TYPE_TO_TYPEINFO.get(schema.type());
-  }
+@Deprecated
+public class HiveSchemaConverter extends io.confluent.connect.storage.hive.HiveSchemaConverter {
```
ewencp: I'm not spotting anywhere this is used in public APIs. Can we just get rid of it and rely on the one in kafka-connect-storage-common?

src/main/java/io/confluent/connect/hdfs/parquet/ParquetFileReader.java
```diff
@@ -18,49 +18,81 @@
...
+    try {
+      ParquetReader<GenericRecord> parquetReader = builder.withConf(conf.getHadoopConfiguration())
+          .build();
+      GenericRecord record;
+      while ((record = parquetReader.read()) != null) {
+        result.add(record);
+      }
+      parquetReader.close();
+    } catch (IOException e) {
+      throw new DataException(e);
```
ewencp: Is `DataException` the right choice here? Couldn't this just be due to connectivity issues with HDFS? My specific concern is that this is not a retriable exception.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -60,7 +61,7 @@ public void append(String tempFile, String committedFile) throws ConnectExceptio
       writer.append(key, value);
       writer.hsync();
     } catch (IOException e) {
-      throw new ConnectException(e);
+      throw new DataException(e);
```
ewencp: Same questions about retriable errors here. I think there are cases where you could fail to acquire the WAL and given the way the state machine works, you'd want a retriable exception. Even if we weren't doing that correctly for all (or any) of these cases, `DataException` seems weird for at least some of the `IOExceptions` that can happen in the `FSWAL` code.

src/test/java/io/confluent/connect/hdfs/FailureRecoveryTest.java
```diff
@@ -29,30 +29,25 @@
...
+    props.put(StorageCommonConfig.STORAGE_CLASS_CONFIG, MemoryStorage.class.getName());
     props.put(HdfsSinkConnectorConfig.FORMAT_CLASS_CONFIG, MemoryFormat.class.getName());
     return props;
   }
 
   @Test
   public void testCommitFailure() throws Exception {
-    Map<String, String> props = createProps();
-    HdfsSinkConnectorConfig connectorConfig = new HdfsSinkConnectorConfig(props);
+    setUp();
```
ewencp: Why did these move inside the tests? It seems like in this test they all happen first in the test method, which seems a good indicator we should just use the `@Before`?

src/test/java/io/confluent/connect/hdfs/TestWithMiniDFSCluster.java
```diff
@@ -63,10 +99,234 @@ private MiniDFSCluster createDFSCluster(Configuration conf) throws IOException {
...
+    for (TopicPartition tp : partitions) {
+      long offset = startOffset;
+      for (Struct record : records) {
+        sinkRecords.add(new SinkRecord(TOPIC, tp.partition(), Schema.STRING_SCHEMA, key, schema, record, offset++));
+      }
+    }
+    return sinkRecords;
+  }
+
+  protected List<SinkRecord> createSinkRecordsNoVersion(int size, long startOffset) {
```
ewencp: Seems like there is probably some opportunity for test utilities in storage-common since a lot of these don't seem dependent on anything in this repo and look very familiar -- it's all the common data generation & validation code.

ewencp:
@kkonstantine Added some merges & updates. Still not fully passing and has some issues. Haven't addressed any review comments, but closer to passing build & tests.

kkonstantine:
This last commit should fix the remaining timebased hive tests when this one is merged: https://github.com/confluentinc/kafka-connect-storage-common/pull/27

However we should consider for compatibility reasons to instantiate a "default" schema generator class, if the user does not supply a schema generator class. Depending on the partitioner, this default might be the default schema generator or the time-based schema generator. Let me know

ewencp:
@kkonstantine I've updated the patch to provide compatibility for `Format` and the classes it effectively makes public, stable API. `Partitioner` looks like it is ok because it maintained the existing interface and only added the storage-common version as an `extends`.

This also fixes up the remaining test issues once https://github.com/confluentinc/kafka-connect-storage-common/pull/31 is merged. I think that rounds this out. We still lack a full end-to-end test with a custom `Format` and `Partitioner` built against the original version, which would be the ideal case, but I think I've done the necessary due diligence to make sure we match the original interfaces (and they are now commented so nobody should modify them in the future) and we have decent unit testing around both old and new cases.

kkonstantine:
One last one (removal of `Override`s from WAL interface) and we are ready for ignition! 

Thanks for the reviews and the commits @ewencp!

PR #192, opened 2017-05-16
Fixed incorrect logic for detecting stale records.

lomignet:
Thanks to @johnhofman in f038915ab00f2bf37fd1882a65d9333dec565d8f.


lomignet:
[clabot:check]

shardnit:
Is it going to master any time soon?

ewencp:
I filed #212 to simply remove this check since it was only there for compatibility with older versions of Kafka, but there have been enough releases and the issue was with the connect framework itself,  so I don't think there's any reason to keep it around anymore. I'm going to close this in favor of that PR, but thank you for the contribution!

PR #194, opened 2017-05-19
Fix HdfsSinkConnector to extend from SinkConnector instead of Connector.

ewencp:


ewencp:
@kkonstantine Trivial fix, same as needed to be done for S3. I've put it all the way back to 3.0.x since  it affects a lot of versions and it should merge easily anyway.

PR #196, opened 2017-05-24
CC-492: Add JsonFormat to HDFS Connector

kkonstantine:


kkonstantine:
Untested

src/main/java/io/confluent/connect/hdfs/json/JsonRecordWriterProvider.java
```diff
@@ -0,0 +1,104 @@
...
+              writer.writeObject(value);
+              writer.writeRaw(LINE_SEPARATOR);
+            }
+          } catch (IOException e) {
+            throw new ConnectException(e);
+          }
+        }
+
+        @Override
+        public void commit() {}
```
rhauch: Is there any advantage to flushing the output stream with `commit()`? `JsonGenerator` doesn't really seem to offer a flush, but `OutputStream` certainly does.

src/main/java/io/confluent/connect/hdfs/json/JsonRecordWriterProvider.java
```diff
@@ -0,0 +1,104 @@
...
+          try {
+            Object value = record.value();
+            if (value instanceof Struct) {
+              byte[] rawJson = converter.fromConnectData(
+                  record.topic(),
+                  record.valueSchema(),
+                  value
+              );
+              out.write(rawJson);
+              out.write(LINE_SEPARATOR_BYTES);
```
rhauch: Why do we generate JSON files with multiple lines?

src/main/java/io/confluent/connect/hdfs/json/JsonFormat.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Map;
+
+import io.confluent.connect.hdfs.HdfsSinkConnectorConfig;
+import io.confluent.connect.hdfs.storage.HdfsStorage;
+import io.confluent.connect.storage.format.Format;
+import io.confluent.connect.storage.format.RecordWriterProvider;
+import io.confluent.connect.storage.format.SchemaFileReader;
+import io.confluent.connect.storage.hive.HiveFactory;
+
+public class JsonFormat implements Format<HdfsSinkConnectorConfig, Path> {
```
rhauch: IIUC, this is the class that users have to specify within the connector's configuration. Wouldn't it be better to have JavaDoc on this class with explanations about how to use it?

kkonstantine: I agree with the necessity of javadocs here and in similar places. 

However when documentation is missing from the older/prexisting similar classes as well, I'm inclined to treat it as tech debt and address it altogether in a separate PR. I find this a reasonable compromise that aims to avoid introducing too much noise in focused PRs and helps not duplicating the effort (write the javadocs once for everything). Of course, there's always the danger to postpone this task a bit too much, but I believe that with the appropriate jiras it could be tracked sufficiently.

rhauch: Wouldn't it be better to not incur more tech debt, especially when the change is fresh in your mind? Seems cheap to do now for this class and expensive to do later.

kkonstantine: I was referring to the absence of javadoc in `AvroFormat` and `ParquetFormat`. I don't mind adding javadoc in `JsonFormat` but I don't think the irrelevant files should be touched as part of this PR.

rhauch: I agree with that: add it here, but not anywhere else.

src/main/java/io/confluent/connect/hdfs/json/JsonRecordWriterProvider.java
kkonstantine: This method is not used in the HDFS connector. It was introduced in the `RecordWriterProvider` interface in `kafka-connect-storage-common` to be used by connectors such as the S3 connector that need to attempt committing offsets before closing the writer.

This connector uses only `close`, which flushes the output stream during `writer.close()` since the JsonGenerator by default manages the underlying stream.

src/main/java/io/confluent/connect/hdfs/json/JsonFormat.java
kkonstantine: I'm a big fan of javadocs. Added a few class-level ones. You might have noticed the wide absence of javadocs in our connectors (storage-common has some). Given that, their addition needs to follow a more systematic approach.

src/main/java/io/confluent/connect/hdfs/json/JsonRecordWriterProvider.java
rhauch: Sounds fine. Thanks for the answer.

kkonstantine:
Thanks @rhauch!

PR #205, opened 2017-06-06
Upgrade avro to 1.8.2

mageshn:


pom.xml
```diff
@@ -55,7 +55,7 @@
         <junit.version>4.12</junit.version>
         <hadoop.version>2.7.3</hadoop.version>
         <hive.version>1.2.1</hive.version>
-        <avro.version>1.7.7</avro.version>
+        <avro.version>1.8.0</avro.version>
```
ewencp: 1.8.2 or, at least 1.8.2 here?

ewencp:
@mageshn Actually, I take it back. This shouldn't be an issue since this is a parquet version and we will already prioritize our avro over theirs. So this LGTM.

PR #217, opened 2017-07-27
Update quickstart to use Confluent CLI

kkonstantine:


docs/hdfs_connector.rst
```diff
@@ -54,16 +82,37 @@ The three records entered are published to the Kafka topic ``test_hdfs`` in Avro
...
 ``etc/kafka-connect-hdfs/quickstart-hdfs.properties`` are properly set to your configurations of
-Hadoop, e.g. ``hdfs.url`` points to the proper HDFS and using FQDN in the host. Then run the
-following command to start Kafka connect with the HDFS connector::
+Hadoop, e.g. ``hdfs.url`` points to the proper HDFS and using FQDN in the host. Then start connector by loading its
+configuration with the following command:
+
+.. sourcecode:: bash
+
+   $ confluent load hdfs-sink -d etc/kafka-connect-hdfs/quickstart-hdfs.properties
+   {
```
rhauch: This line starts the output of the `confluent load` command, right? I know people will think they have to run the whole command including what's in the braces. I suggest adding a normal text line after the command that says something like "This will output the connector configuration that is being started:"

docs/hdfs_connector.rst
```diff
@@ -54,16 +82,37 @@ The three records entered are published to the Kafka topic ``test_hdfs`` in Avro
...
+       "tasks.max": "1",
+       "topics": "test_hdfs",
+       "hdfs.url": "hdfs://localhost:9000",
+       "flush.size": "3",
+       "name": "hdfs-sink"
+     },
+     "tasks": []
+   }
+
+To check that the connector started successfully open the Connect worker's log by running:
```
rhauch: Replace "open" with "view the end of the"?

docs/hdfs_connector.rst
```diff
@@ -54,16 +82,37 @@ The three records entered are published to the Kafka topic ``test_hdfs`` in Avro
...
+.. sourcecode:: bash
 
-  $ ./bin/connect-standalone etc/schema-registry/connect-avro-standalone.properties \
-  etc/kafka-connect-hdfs/quickstart-hdfs.properties
+  $ confluent log connect
 
-You should see that the process starts up and logs some messages, and then exports data from Kafka
-to HDFS. Once the connector finishes ingesting data to HDFS, check that the data is available
-in HDFS::
+By navigating at the end of the log, you should see that the connector starts, logs a few messages, and then exports
```
rhauch: Using "navigating" sounds like they have to find the log file. Perhaps "The last lines in the log should show that the connector starts ...".

docs/hdfs_connector.rst
```diff
@@ -95,13 +150,17 @@ You should see the following output::
 
 .. note:: If you want to run the Quickstart with Hive integration, before starting the connector,
```
rhauch: Seems like this could be a new section, since the quickstart seems like it finished the main demo and is now backtracking to describe what's different when using Hive.

Also, should the quickstart show how to use the CLI to stop the services?

kkonstantine: I agree. Added stop examples in S3's quickstart. I'll add the same here.

There wasn't a section here, and ideally I'd like to restrict the changes to the ones relevant to CLI. Still, after reading again I think the Hive part is pretty small. Not sure it justifies a new (sub)section. Not feeling strong about it either. I'll add it if you prefer it that way.

docs/hdfs_connector.rst
kkonstantine: Will change.

docs/hdfs_connector.rst
kkonstantine: Will change.

docs/hdfs_connector.rst
kkonstantine: I had the same thought. Problem is that we list the output in almost every command. Feels a bit sub-optimal to separate command and output with a sentence everywhere. 

The question is, won't the reader understand what's happening after a couple commands? Pasting the whole thing will generate command line errors for sure. Let me know.

docs/hdfs_connector.rst
rhauch: Not being as familiar with the quickstart, I thought the transition to talk about Hive was a little weird as it stands. Understand that it's not related to the CLI, but it does seem like the transition could be made smoother. It's almost like "If you're using Hive, start over and change the connector config slightly, then see the data in Hive a bit differently." 

Alternatively, perhaps the "If you're using Hive" note could be moved up earlier around when the connector configuration is first discussed. Not sure if that's better or worse.

docs/hdfs_connector.rst
rhauch: People with less experience will do this. Sure, experienced people will read this and feel the steps are too fine grained or the text too verbose. But it seems like a small price if it keeps people from going off into the weeds or getting a bad first impression.

rhauch: The challenge is that inexperienced people are more likely to run the whole thing, and they're also the people that may not really know how to interpret the error. It's up to you.

kkonstantine: To keep uniformity with other quickstarts in other repos, I'm inclined to leave it as is for now, and issue a separate PR. This seems to be a change will small impact. 

I agree with the observation. At the same time someone could argue that all the readers that are ready to use Avro format to ingest their data in our quickstarts won't have a very hard time with this one. Let's reconsider separately soon after the quickstarts are updated.

docs/hdfs_connector.rst
kkonstantine: Again, I'm suggesting addressing this one separately so it won't be hidden under a commit related to the CLI.

kkonstantine:
Thanks @rhauch. Merging this one. Created https://confluentinc.atlassian.net/browse/CC-987 and https://confluentinc.atlassian.net/browse/CC-988 to follow up on the items not addressed here.

PR #220, opened 2017-08-09
HOTFIX: Enable default for format and storage classes in hdfs connector

kkonstantine:


src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -246,7 +247,21 @@
...
+      ConfigDef configDef,
+      Map<String, String> props
+  ) {
+    ConcurrentMap<String, String> propsCopy = new ConcurrentHashMap<>(props);
+    propsCopy.putIfAbsent(StorageCommonConfig.STORAGE_CLASS_CONFIG, HdfsStorage.class.getName());
+    propsCopy.putIfAbsent(HdfsSinkConnectorConfig.FORMAT_CLASS_CONFIG, AvroFormat.class.getName());
+    return new HdfsSinkConnectorConfig(configDef, propsCopy);
+  }
+
+  protected HdfsSinkConnectorConfig(Map<String, String> props) {
```
rhauch: What about keeping the public constructor and instead adding the default properties if missing:

```
public HdfsSinkConnectorConfig(Map<String, String> props) {
    this(CONFIG_DEF, addDefaults(props));
}

private static Map<String, String> addDefaults(Map<String, String> props) {
    ConcurrentMap<String, String> propsCopy = new ConcurrentHashMap<>(props);
    propsCopy.putIfAbsent(StorageCommonConfig.STORAGE_CLASS_CONFIG, HdfsStorage.class.getName());
    propsCopy.putIfAbsent(HdfsSinkConnectorConfig.FORMAT_CLASS_CONFIG, AvroFormat.class.getName());
    return propsCopy;
}
```

The benefit is that the public constructor stays the same, just in case people extend this connector/class. Not sure if that's a worry. But it would reduce the number of changes where the constructor is normally used.

rhauch: Also, this doesn't really fix when validating a configuration via the ConfigDef instance, right?

kkonstantine: Yeah, you reminded me why the voice in my head was telling me _never add a static method in Config again_ ... I'll fix the `public static ConfigDef getConfig() {` which by the way contains a wrong blacklist now.

kkonstantine: Also good point re: the constructor access scope.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -328,17 +340,40 @@ public static ConfigDef getConfig() {
...
+
+    visible.define(
+        STORAGE_CLASS_CONFIG,
+        Type.CLASS,
+        Importance.HIGH,
+        STORAGE_CLASS_DOC,
+        "Storage",
+        1,
+        Width.NONE,
+        STORAGE_CLASS_DISPLAY
```
rhauch: Should these define the default value? If so, then `addDefaults` shouldn't be needed, right?

kkonstantine: That was the purpose :)

kkonstantine: Added. As you can see the challenge is to guess the position of the default value.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -246,8 +251,15 @@
...
+  public static Map<String, String> addDefaults(Map<String, String> props) {
+    ConcurrentMap<String, String> propsCopy = new ConcurrentHashMap<>(props);
+    propsCopy.putIfAbsent(STORAGE_CLASS_CONFIG, HdfsStorage.class.getName());
+    propsCopy.putIfAbsent(HdfsSinkConnectorConfig.FORMAT_CLASS_CONFIG, AvroFormat.class.getName());
+    return propsCopy;
+  }
+
   public HdfsSinkConnectorConfig(Map<String, String> props) {
-    this(CONFIG_DEF, props);
+    this(CONFIG_DEF, addDefaults(props));
```
rhauch: Last comment ... is `addDefaults` necessary anymore since these configurations now have defaults? Otherwise, looks great.

kkonstantine: Glad you mentioned. Reverting the commit that included the changes in the tests removed the test case I added for testing the defaults. 

Yes, it's needed because `getConfig` is static and it's only shadowed by that class. There's no real method override. So even if the base classes use `getConfig` to get the config def, which I don't think they do, the changes in `getConfig` only help for validation.

kkonstantine:
Thanks for the review. Validation was failing on system tests because I didn't add to `skip` the updated properties. Should work now. Will merge once system tests succeed.

kkonstantine:
System tests passed. Merging.

PR #226, opened 2017-09-14
Add upstream project so builds will be triggered automatically

ewencp:


ewencp:
This will work for 3.3.x, but we'll need to also have storage-common as a dependency for 3.4/master

PR #227, opened 2017-09-18
Specify a nodeLabel for Jenkins that doesn't die during the HDFS connector tests.

ewencp:


ewencp:
For reference, this is basically just overriding https://github.com/confluentinc/jenkins-common/blob/master/vars/common.groovy#L13. This probably isn't really ideal unless JDK8 is really required to make the tests work properly, but does address the immediate issue with all merged builds failing.

PR #234, opened 2017-09-29
Addressing commentary from PR #172

cricket007:
See #172 for initial commit. 

Merged and tested against a 3.2.2 release.

### DIFF
--- a/src/test/java/io/confluent/connect/hdfs/HdfsSinkTaskTest.java
+++ b/src/test/java/io/confluent/connect/hdfs/HdfsSinkTaskTest.java
@@ -14,7 +14,6 @@
 
 package io.confluent.connect.hdfs;
 
-import io.confluent.kafka.serializers.NonRecordContainer;
 import org.apache.hadoop.fs.Path;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.data.Schema;
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -402,6 +402,7 @@ public class DataWriter {
   private Map<String, Object> copyConfig(HdfsSinkConnectorConfig config) {
     Map<String, Object> map = new HashMap<>();
     map.put(HdfsSinkConnectorConfig.PARTITION_FIELD_NAME_CONFIG, config.getString(HdfsSinkConnectorConfig.PARTITION_FIELD_NAME_CONFIG));
+    map.put(HdfsSinkConnectorConfig.PARTITION_TIME_FIELD_NAME_CONFIG, config.getString(HdfsSinkConnectorConfig.PARTITION_TIME_FIELD_NAME_CONFIG));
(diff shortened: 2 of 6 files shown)


cricket007:
Ideally, this code should probably be moved up to the common repo. Somehow tied into the other TimeStampExtractors. 

https://github.com/confluentinc/kafka-connect-storage-common/blob/master/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java#L179-L185

kkonstantine:
Also, the target branch should be 3.2.x because `-post` branches are only for documentation changes. Thanks!

src/test/java/io/confluent/connect/hdfs/HdfsSinkTaskTest.java
```diff
@@ -14,7 +14,6 @@
 
 package io.confluent.connect.hdfs;
 
-import io.confluent.kafka.serializers.NonRecordContainer;
```
cricket007: IntelliJ said this wasn't used

cricket007:
@kkonstantine It doesn't seem the common configs support nested timestamps, no. Also, it doesn't seem the HDFS connector is refactored yet to use the common library? At least, that's what I could figure out from at least the 3.3.0 pom. 

I was just following instructions in the other post. 
> We prefer to have PRs for the oldest branches to be changed, and then we merge them onto the newer branches like 3.3.x, 3.4.x and master.

kkonstantine:
Sure, the second part is correct. Hdfs does not use these extractors yet. 

But I'm mentioning because probably a good way to go would be to a) port nested timestamp field extraction to `storage-common` b) refactor hdfs to start using those. 

Ideally, that's how I'd approach it, but I understand this might mean more work. (Also wanted to make sure we are not target 3.2.x for a new feature)

cricket007:
> probably a good way to go would be to a) port nested timestamp field extraction to storage-common 

As far as I can tell, that should not be too hard [looking here](https://github.com/confluentinc/kafka-connect-storage-common/blob/master/partitioner/src/main/java/io/confluent/connect/storage/partitioner/TimeBasedPartitioner.java#L243) to add the loop that was added with this code.

> refactor hdfs to start using those.

Is that effort already underway? I currently working within a 3.2.2 platform, so that's why I chose that branch and have just replaced the connect-hdfs JAR file instead of waiting for this code to be merged upstream. 

I'm willing to change the base branch to whatever it needs to be.

kkonstantine:
Your contribution would be more than welcome. This task is not in progress. 

Since this would be a new functionality it would have to target `master` branch in both repos (Also in order to use `storage-common` from hdfs, you need to work on `master` as well). 

Let us know if you want to work on this. Given the current timelines, it will target the release after the next one (with the next one being 4.0.0, if it's merged on `master` soon, it'll will appear in `4.1.0` - of course being immediately available on `master` for everyone that wants to get it immediately).

Thanks @cricket007 !

cricket007:
@kkonstantine As long as 1) there is nested field support, and 2) the task doesn't end in a halted failure state on a misconfigured RecordField extraction - it can WARN and default to the wall clock, this can be closed in favor of your PR's

cricket007:
Closing in support of confluentinc/kafka-connect-storage-common#48

PR #235, opened 2017-09-29
CC-1172: Fix memory leak in TopicPartitionerWriter

kkonstantine:


rhauch:
@kkonstantine, is it possible with this change that the TopicPartitionWriter performs an offsets.get(partition) and gets null? If so, how does this impact the rest of the connector?

kkonstantine:
I don't see how that's possible. Start offsets and end offsets work in tandem in `TopicPartitionerWriter`. There isn't a single `read` of end offsets that is not preceded by a `read` of a start offset (only exception of an operation on the `offsets` map for which its symmetric does not happen on the `startOffsets`, is a `put` for every new record that is written). 

Plus, after a successful commit, neither a start nor an end offset are needed. And while that means that a removal of the `startOffset` is required in order for the connector to roll files successfully on a - for instance - size based rotation, for `offsets` the absence of removal does not result in error, but it produces the memory leak that is observed if the encoded partition is not used again.

kkonstantine:
Thank you for the reviews!

PR #241, opened 2017-10-04
New parameter to rename partition column

RussWinkler:
Created a new parameter that allows the connector to rename the partition column in the FieldPartitioner.  We found this helps support Presto since two columns with the same name aren't seen as a valid configuration.


RussWinkler:
[clabot:check]

sv-spodina:
@RussWinkler , the feature you implemented helps integration with Hive as well. Looking forward to having the change merged into master as soon as possible.

cricket007:
This partitioner is deprecated as of `4.0.0`, so this should be ported to [the other `FieldPartitioner`](https://github.com/confluentinc/kafka-connect-storage-common/blob/master/partitioner/src/main/java/io/confluent/connect/storage/partitioner/FieldPartitioner.java)

PR #245, opened 2017-10-16
CC-1212: Accept partitioners implementing the new partitioner interface

kkonstantine:

PR #246, opened 2017-10-16
CC-1213: Enable record based timebased partitioners in HDFS connector

kkonstantine:


kkonstantine:
ok to test

kkonstantine:
Aiming to apply minimal changes to `TopicPartitionWriter`'s workflow in order to support timestamp based partitioning (record, or field) and at the same time retain the exact same behavior for wallclock based partitioners. 

This PR is implemented on top of https://github.com/confluentinc/kafka-connect-hdfs/pull/245

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -329,9 +365,6 @@ public void write() {
                 break;
               }
             } else {
-              SinkRecord projectedRecord = compatibility.project(record, null, currentSchema);
```
kkonstantine: This is an important reordering that is required in order to be able to rotate based on a record's timestamp.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -476,7 +501,15 @@ private void setState(State state) {
   }
 
   private boolean shouldRotate(long now) {
-    boolean periodicRotation = rotateIntervalMs > 0 && now - lastRotate >= rotateIntervalMs;
+    currentTimestamp = timestampExtractor.extract(currentRecord);
+    lastRotate = lastRotate == null && currentRecord != null
+                 ? timestampExtractor.extract(currentRecord)
```
rhauch: Do you mean to call `timestampExtractor.extract(currentRecord)` twice within a few lines?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -476,7 +501,15 @@ private void setState(State state) {
   }
 
   private boolean shouldRotate(long now) {
-    boolean periodicRotation = rotateIntervalMs > 0 && now - lastRotate >= rotateIntervalMs;
+    currentTimestamp = timestampExtractor.extract(currentRecord);
+    lastRotate = lastRotate == null && currentRecord != null
```
ewencp: Is this supposed to be updating `lastRotate`? The method name doesn't really indicate it would do so. Perhaps this was supposed to be a local variable?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
kkonstantine: Indeed, but I'm updating the whole block to distinguish between wallclock based and timestamp based partitioning (need to be sure that partitioners that examine the record never operate on `null` records).

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
kkonstantine: This update to the global `lastRotate` needs to be added for timestamp based partitioners for the following reason: 

The only place that `lastRotate` was updated so far was `updateRotationTimers`. However with timestamp based partitioners the case is that in the first call of `updateRotationTimers` and before entering the while loop with `buffer` non-empty, the timestamp-based partitioners don't have a value for `lastRotate`. Then when they enter the loop, they need a call to both: 
```java 
currentTimestamp = timestampExtractor.extract(currentRecord);
lastRotate = lastRotate == null ? currentTimestamp : lastRotate;
```
somewhere, or they'll never rotate based on time. Draining the `buffer` does not make a difference, because the branch on empty `buffer` still calls `shouldRotate` first before calling `updateRotationTimers`. 

Happy to change the name for `shouldRotate` to reflect this change. The reason we don't do this already is probably that we count on processing a batch of records will always take less time than the rotation interval. That's why `now` is passed as argument and is not called every time withing `shouldRotate`. But for timestamp based partitioners this is not an option. Every record needs to be inspected. 

The newly added test case `testWriteRecordTimeBasedPartitionFieldTimestampHours` catches this case. 
(I'm adding some more test cases for record timestamp, but unfortunately a few changes in storage-common are needed too, or a small duplication of code to add `createRecordWithTimestamp`).

kkonstantine:
@ewencp @rhauch Added test cases for timestamp based rotation and schedule based (equivalent to the ones for `kafka-connect-storage-cloud`).

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -206,15 +226,20 @@ public TopicPartitionWriter(
...
-      this.hiveMetaStore = hiveMetaStore;
-      this.hive = hive;
-      this.executorService = executorService;
-      this.hiveUpdateFutures = hiveUpdateFutures;
-      hivePartitions = new HashSet<>();
+    } else {
+      hiveDatabase = null;
     }
 
+    this.hiveMetaStore = hiveMetaStore;
```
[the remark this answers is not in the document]
kkonstantine: I agree. Given the existing constructors I found this to be the less explicit choice. If the simpler constructor is used, they'll be `null` (but that's for tests). But it's on the connector's devs to ensure that none of them is accessed when `hiveIntegration == false`.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -475,8 +505,18 @@ private void setState(State state) {
     this.state = state;
   }
 
-  private boolean shouldRotate(long now) {
-    boolean periodicRotation = rotateIntervalMs > 0 && now - lastRotate >= rotateIntervalMs;
+  private boolean shouldRotateAndMaybeUpdateTimers(long now) {
+    if (isWallclockBased) {
+      currentTimestamp = now;
```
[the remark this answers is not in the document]
kkonstantine: That's the easy one. Changed.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -54,60 +54,68 @@
...
-  private long rotateIntervalMs;
-  private long lastRotate;
-  private long rotateScheduleIntervalMs;
+  private final int flushSize;
+  private final long rotateIntervalMs;
+  private Long lastRotate;
+  private final long rotateScheduleIntervalMs;
   private long nextScheduledRotate;
+  private Long currentTimestamp;
+  private SinkRecord currentRecord;
```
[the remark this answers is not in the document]
kkonstantine: I agree. Probably easier to argue about its usage as local than as a class member, although the effect is the same.

kkonstantine:
I have addressed the comments re: the class variables. 

Also, system tests are passing/expected to pass: 
https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/439
https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/445
https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/446

PR #252, opened 2017-11-03
MINOR: Changed the HDFS configuration to maintain the same group order

rhauch:


src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -371,17 +375,19 @@ public boolean visible(String name, Map<String, Object> connectorConfigs) {
...
     skip.add(FORMAT_CLASS_CONFIG);
 
+    // Order added is important, so that group order is maintained
     ConfigDef visible = new ConfigDef();
+    addAllConfigKeys(visible, newConfigDef(), skip);
+    addAllConfigKeys(visible, StorageCommonConfig.newConfigDef(STORAGE_CLASS_RECOMMENDER), skip);
+    addAllConfigKeys(visible, PartitionerConfig.newConfigDef(STORAGE_CLASS_RECOMMENDER), skip);
+    addAllConfigKeys(visible, HiveConfig.getConfig(), skip);
 
+    // Add the overridden configurations
```
rhauch: @kkonstantine this is the same approach I used in https://github.com/confluentinc/kafka-connect-storage-cloud/pull/118: group ordering is actually insertion order in ConfigDef, so we have to add the groups in the same order. We could use a LinkedHashMap, but we'd have to add the overridden configkeys last, and if we're doing that its simpler to just create the ConfigDef rather than a temp LinkedHashMap.

rhauch:
retest this please

rhauch:
@kkonstantine Fixed the PR so that it passes the build. Please review as soon as you have a chance. Thanks!

PR #255, opened 2017-11-09
CC-1299: Add configuration setting for Avro compression codec

wicknicks:


src/test/java/io/confluent/connect/hdfs/avro/AvroCompressionTest.java
```diff
@@ -0,0 +1,69 @@
...
+      Path p = new Path(s);
+      try (FSDataInputStream stream = fs.open(p)) {
+        int size = (int) fs.getFileStatus(p).getLen();
+        ByteBuffer buffer = ByteBuffer.allocate(size);
+        if (stream.read(buffer) <= 0) {
+          log.error("Could not read file {}", s);
+        }
+
+        String deserialized = new String(buffer.array());
+        int ix;
```
kkonstantine: For the same reason as above, acronyms are tricky too. Would just `i` make sense here? (When I show it I was expecting to find `i` in an outer loop or something. 

A counter example of a sensible acronym is `fs` above. I guess because it's used in writing too more often.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -196,6 +203,16 @@ public static ConfigDef newConfigDef() {
...
           LOGS_DIR_DISPLAY
       );
+
+      // define avro codec
+      configDef.define(AVRO_CODEC_CONFIG,
+          Type.STRING,
+          AVRO_CODEC_DEFAULT,
+          Importance.LOW,
+          AVRO_CODEC_DOC,
+          group, ++orderInGroup, Width.LONG,
```
kkonstantine: In the current code style (outside AK) we use one argument per line in multi-line method calls. A bit wasteful in terms of space but easier to maintain and not miss an argument.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -79,6 +79,13 @@
   public static final String LOGS_DIR_DEFAULT = "logs";
   public static final String LOGS_DIR_DISPLAY = "Logs directory";
 
+  public static final String AVRO_CODEC_CONFIG = "avro.codec";
+  public static final String AVRO_CODEC_DEFAULT = "null";
+  public static final String AVRO_CODEC_DISPLAY = "Avro Compression Codec";
+  public static final String AVRO_CODEC_DOC = "The Avro compression codec to be used for "
+      + "output files. Available values: null, deflate, "
```
kkonstantine: nit: Line width limit is 100. Maybe linebreaks need adjustment here (still 3 lines probably)

norwood:
retest this please

src/test/java/io/confluent/connect/hdfs/avro/AvroCompressionTest.java
wicknicks: I could use `index` here. As you said, `i` is usually used as a loop variable. `index` would be relevant to what it contains too.

src/main/java/io/confluent/connect/hdfs/avro/AvroRecordWriterProvider.java
```diff
@@ -56,6 +57,11 @@ public String getExtension() {
       final Path path = new Path(filename);
       Schema schema = null;
 
+      public io.confluent.connect.storage.format.RecordWriter init() {
```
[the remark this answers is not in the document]
wicknicks: Great find! Thanks! Yeah, that can help get rid of the init method.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
wicknicks: The lines here are at most 91 characters long. I think we should be fine with respect to the 100 line limit?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -280,6 +235,76 @@ public static ConfigDef newConfigDef() {
     return configDef;
   }
 
+  /**
+   * Define HDFS configuration group
+   */
+  private static void newHadoopDef(ConfigDef configDef) {
```
wicknicks: moved these statements into a separate method because we have a checkstyle requirement of methods being at most 150 lines.

src/main/java/io/confluent/connect/hdfs/avro/AvroRecordWriterProvider.java
```diff
@@ -71,7 +73,7 @@ public void write(SinkRecord record) {
           }
         }
 
-        log.trace("Sink record: {}", record.toString());
+        log.trace("Sink record: {}", record);
```
wicknicks: removed the .toString() method call because it is redundant and log level will be higher than Trace in most situations (we can save some CPU cycles).

src/main/java/io/confluent/connect/hdfs/avro/AvroRecordWriterProvider.java
```diff
@@ -52,7 +53,8 @@ public String getExtension() {
       final String filename
   ) {
     return new io.confluent.connect.storage.format.RecordWriter() {
-      final DataFileWriter<Object> writer = new DataFileWriter<>(new GenericDatumWriter<>());
+      final DataFileWriter<Object> writer = new DataFileWriter<>(new GenericDatumWriter<>())
+          .setCodec(CodecFactory.fromString(conf.getAvroCodec()));
```
rhauch: What happens when `conf.getAvroCodec()` returns null?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -79,6 +79,13 @@
   public static final String LOGS_DIR_DEFAULT = "logs";
   public static final String LOGS_DIR_DISPLAY = "Logs directory";
 
+  public static final String AVRO_CODEC_CONFIG = "avro.codec";
+  public static final String AVRO_CODEC_DEFAULT = "null";
+  public static final String AVRO_CODEC_DISPLAY = "Avro Compression Codec";
+  public static final String AVRO_CODEC_DOC = "The Avro compression codec to be used for output  "
+      + "files. Available values: null, deflate, snappy and bzip2 (CodecSource is org.apache"
+      + ".avro.file.CodecFactory)";
```
rhauch: Should we provide a validator for this config?

src/main/java/io/confluent/connect/hdfs/avro/AvroRecordWriterProvider.java
kkonstantine: Makes sense, `sfl4j` implementation will call this anyway.

src/main/java/io/confluent/connect/hdfs/avro/AvroRecordWriterProvider.java
wicknicks: If conf.getAvroCodec() returns a "null" (a String), then CodecFactory returns a No-op codec. This is the default codec used by Avro. If the function returns a null, or an unknown codec (lz4, for example), we get an AvroRuntimeException.

kkonstantine: This can't happen given that parsing of String values is required by `AbstractConfig`. That's why the default can't be set to `null` and as in the S3 it's instead set to `"null"`.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
kkonstantine: Seems like if another config is added the new method will suffer from the same issue. For things like that, adding an as specific as possible suppression in checkstyle might be reasonable as well.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
wicknicks: Added a validator here.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -280,6 +237,77 @@ public static ConfigDef newConfigDef() {
...
+    // define avro codec
+    configDef.define(AVRO_CODEC_CONFIG,
+        Type.STRING,
+        AVRO_CODEC_DEFAULT,
+        ConfigDef.ValidString.in(AVRO_SUPPORTED_CODECS),
+        Importance.LOW,
+        AVRO_CODEC_DOC,
+        group,
+        ++orderInGroup,
+        Width.LONG,
```
rhauch: Does this need to be `Width.LONG`? Seems like the values are all relatively short.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -273,7 +276,10 @@ public static ConfigDef newConfigDef() {
       );
     }
     // Put the storage group(s) last ...
-    ConfigDef storageConfigDef = StorageSinkConnectorConfig.newConfigDef(FORMAT_CLASS_RECOMMENDER);
+    ConfigDef storageConfigDef = StorageSinkConnectorConfig.newConfigDef(
+        FORMAT_CLASS_RECOMMENDER,
+        AVRO_COMPRESSION_RECOMMENDER);
```
kkonstantine:
```java 
ConfigDef storageConfigDef = StorageSinkConnectorConfig.newConfigDef(
        FORMAT_CLASS_RECOMMENDER,
        AVRO_COMPRESSION_RECOMMENDER
);
```

src/test/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfigTest.java
```diff
@@ -68,6 +71,25 @@ public void testUndefinedURL() throws Exception {
...
+    }
+  }
+
+  @Test(expected = ConfigException.class)
+  public void testUnsupportedAvroCompressionSettings() {
+    // test for an unsupported codec.
+    this.properties.put(HdfsSinkConnectorConfig.AVRO_CODEC_CONFIG, "abc");
+
+    new HdfsSinkConnectorConfig(properties);
+    Assert.assertTrue("Expected the constructor to throw an exception",false);
```
kkonstantine: Formatting will show that there's a missing whitespace character after comma in `exception",false);`

wicknicks: Thanks, @kkonstantine. Added the whitespace.

wicknicks:
retest this please

PR #256, opened 2017-11-23
Should consider the situation that late message

asdf2014:
As far as I know, the latest `shouldRotateAndMaybeUpdateTimers` method only considers messages that are received prematurely. Maybe we should handle those messages that are received late.

### DIFF
--- a/src/test/java/io/confluent/connect/hdfs/avro/TopicPartitionWriterTest.java
+++ b/src/test/java/io/confluent/connect/hdfs/avro/TopicPartitionWriterTest.java
@@ -344,6 +344,94 @@ public class TopicPartitionWriterTest extends TestWithMiniDFSCluster {
   }
 
   @Test
+  public void testWriteRecordTimeBasedPartitionFieldTimestampHoursLateMessage() throws Exception {
+    // Do not roll on size, only based on time.
+    localProps.put(HdfsSinkConnectorConfig.FLUSH_SIZE_CONFIG, "1000");
+    localProps.put(
+        HdfsSinkConnectorConfig.ROTATE_INTERVAL_MS_CONFIG,
+        String.valueOf(TimeUnit.MINUTES.toMillis(1))
+    );
+    setUp();
+
+    // Define the partitioner
+    partitioner = new DataWriter.PartitionerWrapper(new HourlyPartitioner<FieldSchema>());
+    parsedConfig.put(PartitionerConfig.TIMESTAMP_EXTRACTOR_CLASS_CONFIG, "RecordField");
+    partitioner.configure(parsedConfig);
+
+    TopicPartitionWriter topicPartitionWriter = new TopicPartitionWriter(
+        TOPIC_PARTITION,
+        storage,
+        writerProvider,
+        newWriterProvider,
+        partitioner,
+        connectorConfig,
+        context,
+        avroData,
+        time
+    );
+
+    Schema schema = createSchemaWithTimestampField();
+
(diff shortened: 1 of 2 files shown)


asdf2014:
[clabot:check]

asdf2014:
Oops, some problems in CI process. But why i cannot visit the [jenkins](https://jenkins.confluent.io/job/confluentinc-pr/job/kafka-connect-hdfs/job/PR-256/1/display/redirect) url...

kkonstantine:
@asdf2014 jenkins does not offer anonymous access here. 

Your unit test failed.

asdf2014:
@kkonstantine Hi, thank you for your comment.  I wonder where can i get the `1.1.0-SNAPSHOT` version kafka. I found the newest version of kafka is `1.0.0` in maven center [repository](https://mvnrepository.com/). When i upgraded the Kafka Connect from `3.3.1` into `4.1.0-SNAPSHOT`, i couldn't run those unittest cases on local due to dependecy kafka that version is `1.1.0-SNAPSHOT`...

kkonstantine:
@asdf2014 to build `1.1.0-SNAPSHOT` checkout and install locally the code from `trunk` in https://github.com/apache/kafka

For an (unsorted) list of the rest of the dependencies you'll need to build when working with snapshot versions of this connector, check the FAQ here: https://github.com/confluentinc/kafka-connect-hdfs/wiki/FAQ

src/test/java/io/confluent/connect/hdfs/avro/TopicPartitionWriterTest.java
```diff
@@ -343,6 +343,94 @@ public void testWriteRecordTimeBasedPartitionFieldTimestampHours() throws Except
...
+    }
+    sinkRecords.addAll(createSinkRecords(
+        records.subList(9, 18),
+        schema,
+        9,
+        Collections.singleton(new TopicPartition(TOPIC, PARTITION))
+    ));
+
+    // And one last record to flush the previous ones.
+    long timestampMuchLater = first.minusHours(6).getMillis();
```
cyrusv: Later or earlier?

src/test/java/io/confluent/connect/hdfs/avro/TopicPartitionWriterTest.java
```diff
@@ -343,6 +343,94 @@ public void testWriteRecordTimeBasedPartitionFieldTimestampHours() throws Except
...
+    int size = 18;
+
+    ArrayList<Struct> records = new ArrayList<>(size);
+    for (int i = 0; i < size / 2; ++i) {
+      records.add(createRecordWithTimestampField(schema, timestampFirst));
+      timestampFirst += advanceMs;
+    }
+    Collection<SinkRecord> sinkRecords = createSinkRecords(records.subList(0, 9), schema);
+
+    long timestampLater = first.minusHours(2).getMillis();
```
cyrusv: Later or earlier?

cyrusv:
@asdf2014 I understand the change, but I believe this is an unnecessary change. While it is non-breaking, it does open the door to committing many small files in the case of jittered incoming timestamps.

The reason I don't see current behavior as a problem is because late-arriving data in this case will still be put into the appropriate encoded partition writer, so there will be no data loss and data is still going to the correct writer.

To explain the asymmetry in the current code, rotating on events after the "rotateIntervalMs" is required because that's the definition of that configuration option. As far as I can tell, there is no expectation that the connector should trigger additional rotations in the event of late data.

PR #257, opened 2017-11-24
Remove redundant / from tmp dir and other improvements about HdfsSinkConnectorConstants

asdf2014:
Remove redundant / from tmp dir


asdf2014:
@kkonstantine PTAL

PR #259, opened 2017-11-28
Add commons-lang3, kyro and avro-mapred as dependencies

wicknicks:
We are removing storage-common's dependency on hive-exec and instead using
hive-exec:core. This means the HDFS connector has to include the missing
libraries itself. Here we are adding commons-lang3, kyro and avro-mapred.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -105,6 +105,21 @@
             <version>${confluent.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.esotericsoftware.kryo</groupId>
+            <artifactId>kryo</artifactId>
+            <version>2.22</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro-mapred</artifactId>
+            <version>1.8.2</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-minicluster</artifactId>
             <version>${hadoop.version}</version>


pom.xml
```diff
@@ -104,6 +104,21 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.esotericsoftware.kryo</groupId>
+            <artifactId>kryo</artifactId>
+            <version>2.22</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro-mapred</artifactId>
+            <version>1.8.2</version>
```
rhauch: Should we use Maven properties for these versions?

wicknicks: yes, good point. since these are exclusively here to support hive, should we prefix the variables with `hive-` (for example, hive-lang3.version)?

rhauch: How about `${hive.commons.lang3.version}` and `${hive.kyro.version}`, and then `${avro.version}` which is the same that [Schema Registry uses]() but which isn't in any of the parent POMs for this repo. 

So, we can either duplicate `${avro.version}` here (risking divergence from SR), or we can instead put the `${avro.version}` Maven property in the common parent POM and have Schema Registry and this repo both inherit that property value. Any thoughts, @ewencp and @kkonstantine?

wicknicks: I would prefer to reuse avro.version as well, but it doesn't look like we can inherit it as we depend on avro-connector, and not schema registry. Any thoughts on what can be done here?

rhauch:
This change is a result of a [PR on Storage Common repo](https://github.com/confluentinc/kafka-connect-storage-common/pull/49).

pom.xml
rhauch: We inherit properties via parent POMs, which means we could change [Common's POM](https://github.com/confluentinc/common/blob/master/pom.xml) to add the property, and then we would inherit it since [Storage Common inherits from Common](https://github.com/confluentinc/kafka-connect-storage-common/blob/master/pom.xml#L25-L27). Then, the [Schema Registry parent POM](https://github.com/confluentinc/schema-registry/blob/master/pom.xml) would also inherit it (via REST Utils that inherits from Common), and the SR and Avro Converters all get it that way as well.

rhauch:
@wicknicks, actually, I wonder if this should be made on an older branch. Remember that unlike AK, we issue PRs against the oldest branch we plan to update.

wicknicks:
@ewencp: good question. I put them here because only hdfs-connector invokes hive in such a way to require those libraries (s3 for example, does not). Also, we could put them in storage-common to put all hive (and friends) in one place.

pom.xml
kkonstantine: I'm not enthusiastic w.r.t prefixing with the package that is requiring this dependency. 

This information can and should be extracted reliably via 
`mvn dependency:tree -Dverbose`

If another package needs the same dependency in the future, either the prefix will seem strange or another property with a prefix will be required. This doesn't sound ideal.

wicknicks:
OK. I'll close this PR and move the changes to storage-common. Will make another PR for backporting.

pom.xml
wicknicks: good point, @kkonstantine. But the dependency tree will not show a dependency between hive and kryo (for example) because hive-exec doesn't explicitly declare a dependency to kryo (that's why we need to add this ourselves). If you look over the dependencies, kryo and commons-lang will look like its not being used anywhere in the code (and might get incorrectly removed).

PR #269, opened 2018-01-03
Fix or suppress findbugs errors

ewencp:


kkonstantine:
I approved, but where do you expect to have the declaration of the findbugs dependency. In `common`, `kafka-connect-storage-common` or here? Current build complains for missing version.

ewencp:
I have https://github.com/confluentinc/common/pull/87 outstanding, but am not merging until I've fixed all the downstream issues that would cause test failures.

PR #275, opened 2018-01-20
MINOR: Add storage-common as dependency/requirement

kkonstantine:


kkonstantine:
Thanks! Merging after successfully redirecting pre version 4 packaging to branches `rpm-pre-v4` and `debian-pre-v4` respectively. For versions 4 and beyond packaging for this connector still tracks `rpm` and `debian` branches.

PR #284, opened 2018-02-02
modify COMMMITTED_FILENAME_SEPARATOR to under line

yudong2015:
If filename in HDFS with plus(+) , it can not be viewed at HUE. For example, if filename is **000010+000020.avro** , then you can not get file from brower caused by not found. The plus will be unquoted to space at backend, so you can not get the file with plus.

dingtalk20180202104939

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/FileUtils.java
+++ b/src/main/java/io/confluent/connect/hdfs/FileUtils.java
@@ -82,17 +82,18 @@ public class FileUtils {
       long startOffset,
       long endOffset,
       String extension,
-      String zeroPadFormat
+      String zeroPadFormat,
+      String fileDelim
   ) {
     String topic = topicPart.topic();
     int partition = topicPart.partition();
     StringBuilder sb = new StringBuilder();
     sb.append(topic);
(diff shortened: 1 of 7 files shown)


cricket007:
See `file.delim` 

https://docs.confluent.io/current/connect/connect-hdfs/docs/configuration_options.html#storage

kkonstantine:
@cricket007 as I mentioned above, code changes are required to enable this config option to the HDFS connector. @yudong2015 I'll wait a bit in case you want to apply those changes. Otherwise I'll close this PR and open another one.

cricket007:
Oh, thought the property was already available since it's part of the documentation. Must be parsed out from the storage common project

kkonstantine:
Right, but I had commented on that already above. Enabling it is preferable compared to removing it from the docs.

cricket007:
Sorry, I meant the property was _used_, not just "available"

kiuby88:
Which is the status of this PR?
I am trying to replace `+` in the generated file names by `-`, but I do not find a solution

yudong2015:
@kiuby88 I will modify this commit, and try to PR again.

PR #293, opened 2018-03-07
Explicitly specify Jackson version instead of relying on transitive dependency version.

ewencp:


ewencp:
Depends on https://github.com/confluentinc/common/pull/102 being merged first

PR #299, opened 2018-03-19
CC-1669: Set no offset when no offset was found for topic partition

rhauch:
The HDFS connector always uses an offset of 0 when resetting offsets and there is no output in HDFS for the requested topic partition. This results in the connector _always_ starting at the beginning, regardless of the `consumer.auto.offset.reset` setting.

This commit changes the behavior to not explicitly set an offset when the connector found no output in HDFS for the requested topic partition. As a result, the connector will rely upon the consumer to start where it has been configured, either the beginning if `consumer.auto.offset.reset=earliest` (which is what Connect explicitly sets by default) or at the most recent offset if `consumer.auto.offset.reset=latest` when the worker is explicitly configured this way.

Incidentally, this new behavior matches that of the S3 connector.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -618,9 +618,15 @@ public class TopicPartitionWriter {
       // tempfile, but then that tempfile was discarded). To protect against this, even if we
       // just want to start at offset 0 or reset to the earliest offset, we specify that
       // explicitly to forcibly override any committed offsets.
-      long seekOffset = offset > 0 ? offset : 0;
-      log.debug("Resetting offset for {} to {}", tp, seekOffset);
-      context.offset(tp, seekOffset);
+      if (offset > 0) {
+        long seekOffset = offset;
+        log.debug("Resetting offset for {} to {}", tp, seekOffset);
+        context.offset(tp, seekOffset);
+      } else {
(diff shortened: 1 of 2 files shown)


rhauch:
@kkonstantine, @ewencp: in addition to reviewing the PR specifics, can you please confirm that resetting offsets should only not find an offset in HDFS for a topic partition if there is no data in HDFS for that topic partition? This could happen upon initial startup of the connector, or if the connector is restarted after a user added more topics to the connector's configuration or if the topic is repartitioned.

ewencp:
@rhauch Yes, that's correct, and the updated behavior seems like what we'd actually want.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -618,9 +618,15 @@ private void resetOffsets() throws ConnectException {
       // tempfile, but then that tempfile was discarded). To protect against this, even if we
       // just want to start at offset 0 or reset to the earliest offset, we specify that
       // explicitly to forcibly override any committed offsets.
-      long seekOffset = offset > 0 ? offset : 0;
-      log.debug("Resetting offset for {} to {}", tp, seekOffset);
-      context.offset(tp, seekOffset);
+      if (offset > 0) {
+        long seekOffset = offset;
```
kkonstantine: Now, with this refactoring, seems safe to just use `offset` (it's still a primitive) in this branch. I think it's simpler than declaring `seekOffset`. WDYT?

PR #300, opened 2018-03-20
Make confluent CLI command output shared

joel-hamill:
Made this an include, so that an easy update going forward. See source: https://github.com/confluentinc/docs/blob/master/includes/installation.rst


joel-hamill:
@kkonstantine yes, definitely will include control-center (sorry about that). here's the source that the include is pointing at: https://github.com/confluentinc/docs/blob/master/includes/installation.rst

PR #304, opened 2018-03-23
CC-1700: Revert commits for CC-1669

wicknicks:
This caused a system test to fail. We believe that not explicitly setting the consumer offset in the consumer allows it to pick up some incorrect value from `__consumer_offsets`, and continue processing from there on. This explains the missing data that was observed by the system.

Please find a clean run on Jenkins here: http://confluent-platform-branch-builder-system-test-results.s3-us-west-2.amazonaws.com/2018-03-23--002.1521777128--wicknicks--HEAD--7138f66/report.html


wicknicks:
Closing in favor of https://github.com/confluentinc/kafka-connect-hdfs/pull/305.

PR #305, opened 2018-03-23
CC-1669: Disable commit of consumer offsets

wicknicks:
This change allows the HDFS connector to tell Connect that consumer offsets need not be committed. 

When a new HDFS connector is started, no files are in HDFS for these topic partitions and the connector will then delegate to the consumer as to where to start. Since this PR changes the connector to have no offsets committed to `__consumer_offsets`, there will be no offsets and the consumer will start based upon the `consumer.auto.offset.reset` Connect worker setting (which Connect defaults to `earliest`), per #299.

Note: If any `__consumer_offsets` for the consumer group used by this new connector _are_ found,
 the consumer will use them. This does allow a user to use the Kafka offset reset tool _before_ the connector is start to control exactly where they want the connector to begin.

However, once the HDFS connector writes files to HDFS for each topic partition, then upon startup the connector will always begin at the next offset past the last one written for that topic (via `context.offsets`, causing Connect will seek the connector to that offset).

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -25,6 +25,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 
@@ -108,9 +109,16 @@ public class HdfsSinkTask extends SinkTask {
   }
 
   @Override
+  public Map<TopicPartition, OffsetAndMetadata> preCommit(
+      Map<TopicPartition, OffsetAndMetadata> currentOffsets
+  ) {
+    // return an empty map so Connect doesn't commit offsets
+    return Collections.emptyMap();
+  }
+
+  @Override
   public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
-    // clear the offsets here so we don't commit to __consumer_offsets.
-    offsets.clear();
+    // Do nothing as the connector manages the offset
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -93,6 +93,9 @@ public class HdfsSinkTask extends SinkTask {
         hdfsWriter.stop();
       }
     }
+
+    log.info("Consumer offsets will be obtained from filenames on HDFS. This connector will not "
+        + "commit or use any offsets from the __consumer_offsets topic.");
   }
 
   @Override
@@ -106,7 +109,8 @@ public class HdfsSinkTask extends SinkTask {
 
   @Override
   public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
-    // Do nothing as the connector manages the offset
+    // clear the offsets here so we don't commit to __consumer_offsets.
+    offsets.clear();
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -95,8 +95,12 @@ public class HdfsSinkTask extends SinkTask {
       }
     }
 
-    log.info("Consumer offsets will be obtained from filenames on HDFS. This connector will not "
-        + "commit or use any offsets from the __consumer_offsets topic.");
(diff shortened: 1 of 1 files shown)


rhauch:
@kkonstantine thanks for the feedback. @wicknicks and I talked about removing `flush` and adding `preCommit` but were trying to have minimal changes, and it so happens that clearing the offsets map happens to work. 

However, I agree with your comment and that adding `preCommit` is better. We can then remove the `flush` call since it's not needed anymore (and the default is no-op anyway).

wicknicks:
@rhauch, @kkonstantine: I've moved the change to `preCommit`. It seems to be a better fit (given the contract mentioned in the SinkTask documentation). Also, we got a clean Jenkins run with the previous change [here](https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/701/). I'll kick off another run, so we can make sure this change is also good.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -93,6 +94,9 @@ public void start(Map<String, String> props) {
         hdfsWriter.stop();
       }
     }
+
+    log.info("Consumer offsets will be obtained from filenames on HDFS. This connector will not "
+        + "commit or use any offsets from the __consumer_offsets topic.");
```
kkonstantine: nit: the ``or use`` part is not accurate. The connector will use consumer offsets from that topic iff files are not present in hdfs and such offset exists for this consumer group. 

``This connector will not commit any offsets to the '__consumer_offsets' topic.`` might be what we want.

rhauch: Yeah, I like @kkonstantine's suggestion. @wicknicks, it's a minor change but would be nice to have.

wicknicks: So we deliberately put the "use any offsets" part to simplify the understanding on where the offsets come from. Apparently, offsets management in HDFS has been confusing in the past. @rhauch?

wicknicks:
Also, looks like we a clean test run here: https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/706/.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -93,6 +94,13 @@ public void start(Map<String, String> props) {
         hdfsWriter.stop();
       }
     }
+
+    log.info("HDFS connector does not commit consumer offsets to Kafka. Upon startup, HDFS "
+        + "Connector restores offsets from filenames in HDFS. In the absence of files in HDFS, "
+        + "the connector will attempt to find an offset for its consumer group in the "
+        + "'__consumer_offsets' topic. If no offsets are found in this topic, the consumer will "
```
kkonstantine: since plural is not accurate here anyways (`If no offsets are found` - the offset is one) should we make it shorter? `If none is found, the consumer will rely ... ` (or `If no offset is found`)

wicknicks: aren't there offsets per partition?

jtuberville:
This was unexpected. Part of our monitoring was depending on consumer offsets.  While this is backwards compatible, seems like more significant than a patch release.

wicknicks:
using the `__consumer_offsets` topic in the HDFS connector can give incorrect results. Since the committed offsets are determined by the committed files in HDFS, the value in `__consumer_offsets` is unreliable. Unfortunately, we don't have expose the committed offset through a metric right now. 

@clabie: there is a consumer lag metric exposed by every KafkaConsumer through JMX. Look at the attribute `records-lag-max` in `kafka.consumer:type=consumer-fetch-manager-metrics` JMX metrics. Maybe you can use this instead of what is in __consumer_offsets to monitor lag?

@jtuberville: would the above help with your use-case? let us know if it doesn't work. thanks.

clabie:
@wicknicks , let me provide you some context

Initially when looking for a monitor solution of kafka consumer lag we considered exposing these JMX metrics for monitoring purposes. However we quickly realised that the lag metric isn't sufficient and needs further interpretation to report a correct status as lag can fluctuate and therefore impossible to define a fixed threshold.

After reading some articles on monitoring consumer lag we came across a [**tool** ](https://github.com/linkedin/Burrow) that offered a complete set of features to monitor consumer lag, including interpretation of lag over a sliding window. For us this was a perfect match as we didn't need to build the entire monitoring framework ourselves. As you can guess by now, it uses the __consumer_offets to get the lag metrics ;)

PR #307, opened 2018-03-30
CC-918 Add Kafka Connect Maven plugin to build

C0urante:
**IMPORTANT**: This PR is targeted at `master` for the moment but should be redirected to `4.1.x` **after** the release occurs. DO NOT MERGE UNTIL THEN

Edit: We've now decided to redirect to `4.0.x`, and will be merging after the release is officially announced.

This PR adds the Kafka Connect Maven plugin to the project's build. Running `mvn package` should cause a plugin archive file to be generated in the `target/plugins/packages` directory, named `confluentinc-kafka-connect-hdfs-<version>-plugin.zip`.


pom.xml
```diff
@@ -144,6 +144,57 @@
...
+                <version>0.9.0-SNAPSHOT</version>
+                <artifactId>kafka-connect-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>kafka-connect</goal>
+                        </goals>
+                        <configuration>
+                            <title>Kafka Connect HDFS</title>
+                            <logo>logos/hdfs.jpeg</logo>
```
rhauch: Might be good to document in the PR the terms that allow us to reuse the logo.

pom.xml
```diff
@@ -144,6 +144,57 @@
...
+
+                            <tags>
+                                <tag>hadoop</tag>
+                                <tag>hdfs</tag>
+                                <tag>hive</tag>
+                            </tags>
+
+                            <requirements>
+                                <requirement>Confluent Platform installed and running</requirement>
+                                <requirement>Hadoop installed and running</requirement>
```
rhauch: Probably sufficient to just specify the Hadoop versions rather than say "... installed and running".

pom.xml
```diff
@@ -144,6 +144,57 @@
...
+                            </pluginTypes>
+
+                            <tags>
+                                <tag>hadoop</tag>
+                                <tag>hdfs</tag>
+                                <tag>hive</tag>
+                            </tags>
+
+                            <requirements>
+                                <requirement>Confluent Platform installed and running</requirement>
```
rhauch: Shouldn't a user be able to install this with just AK?

pom.xml
C0urante: Couldn't find anything hard on what versions of Hadoop the connector works with (I know we have `2.7.3` specified in the POM but not sure about the implications for forwards/backwards compatibility with other versions). Unless we can provide a useful version range here it might make more sense to just eliminate the requirement altogether, since just specifying `Hadoop` seems obvious.

PR #330, opened 2018-05-23
CC-2043: Enable all lint options besides deprecation for source and tests with Java 8

kkonstantine:


kkonstantine:
Depends on: https://github.com/confluentinc/kafka-connect-storage-common/pull/68

kkonstantine:
Thanks. Indeed, deprecation warnings are disabled temporarily because most of them are caused by our own deprecation of interfaces after moving to `kafka-connect-storage-common`

Issue #343, opened 2018-07-10
Isn't this trace a performance killer?


PascalChardon:
https://github.com/confluentinc/kafka-connect-hdfs/blob/3ed231552241d278d5020ff497cd25971699a0ab/src/main/java/io/confluent/connect/hdfs/parquet/ParquetRecordWriterProvider.java#L82

The call to toString method is systematic whatever is the log level.
I've not tried by this way should be better:
log.trace("Sink record: {}", record)

Pascal

kkonstantine:
Thanks @PascalChardon !
The fix makes sense. I consider your fix a performance bug and I'm rebasing your PR on top of `4.0.x` and porting the fix to the avro provider which contains the original trace line that was ported here too in the first place. Will push the additional commit to your branch.

kkonstantine:
Closing as fixed.

PR #346, opened 2018-07-12
Fixed issue #343

PascalChardon:
Small fix preventing the call to toString on each written record in parquet format. The purpose is to enhance performances when logging trace level is disabled.
Issue described in #343 .


PascalChardon:
[clabot:check]

kkonstantine:
Thanks @PascalChardon !
The fix makes sense. I consider your fix a performance bug and I'm rebasing your PR on top of 4.0.x and porting the fix to the avro provider which contains the original trace line that was ported here too in the first place. Will push the additional commit to your branch.

Issue #53, opened 2016-05-04
Recovery persistently fails due to empty WAL file  


vikeshkhanna:
We noticed that two partitions of one of our topics had an ever increasing offset lag. This was happening because Kafka Connect persistently failed to recover these two partitions. We get the following error every time in the log4j files:

```
[2016-05-04 17:30:02,865] ERROR Recovery failed at state RECOVERY_PARTITION_PAUSED (io.confluent.connect.hdfs.TopicPartitionWriter:221)
org.apache.kafka.connect.errors.ConnectException: Error creating writer for log file hdfs://bikeshed//kafka_connect/logs/goscribe.mp-hash_events/48/log
    at io.confluent.connect.hdfs.wal.FSWAL.acquireLease(FSWAL.java:91)
    at io.confluent.connect.hdfs.wal.FSWAL.apply(FSWAL.java:105)
    at io.confluent.connect.hdfs.TopicPartitionWriter.applyWAL(TopicPartitionWriter.java:448)
    at io.confluent.connect.hdfs.TopicPartitionWriter.recover(TopicPartitionWriter.java:204)
    at io.confluent.connect.hdfs.DataWriter.recover(DataWriter.java:236)
    at io.confluent.connect.hdfs.DataWriter.onPartitionsAssigned(DataWriter.java:299)
    at io.confluent.connect.hdfs.HdfsSinkTask.onPartitionsAssigned(HdfsSinkTask.java:103)
    at org.apache.kafka.connect.runtime.WorkerSinkTask$HandleRebalance.onPartitionsAssigned(WorkerSinkTask.java:362)
    at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.onJoinComplete(ConsumerCoordinator.java:194)
    at org.apache.kafka.clients.consumer.internals.AbstractCoordinator.ensureActiveGroup(AbstractCoordinator.java:225)
    at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.ensurePartitionAssignment(ConsumerCoordinator.java:311)
    at org.apache.kafka.clients.consumer.KafkaConsumer.pollOnce(KafkaConsumer.java:890)
    at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:853)
    at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:171)
[...]

PR #348, opened 2018-07-17
CC-274: Recover cleanly with empty WAL file

C0urante:
Addresses #53
[JIRA](https://confluentinc.atlassian.net/browse/CC-274)

If the connector is killed in between creating a new log file and writing a version header to that file, it then ends up in an endless loop during the recover process where it continually tries and fails to read a version header from that file.

This fix involves performing a rudimentary check to see if the WAL file contains an empty/incomplete version header, and if so, creating a new WAL file and overwriting the existing one.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
+++ b/src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
@@ -23,6 +23,7 @@ import org.apache.hadoop.fs.ChecksumException;
 import org.apache.hadoop.fs.CommonConfigurationKeys;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.Syncable;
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -208,8 +210,11 @@ public static Option blockSize(long value) {
         long blockSize = blockSizeOption == null ? fs.getDefaultBlockSize(p) :
                          blockSizeOption.getValue();
 
-        if (appendIfExistsOption != null && appendIfExistsOption.getValue()
-            && fs.exists(p)) {
+        if (appendIfExistsOption != null
+            && appendIfExistsOption.getValue()
+            && fs.exists(p)
+            && hasIntactVersionHeader(p, fs)
```
ewencp: I think if `hasIntactVersionHeader` fails, we probably want to log some error level message. As discussed elsewhere, the fact that we hit that condition should be pretty big red flag -- either Connect is randomly dying mid-write or somethings up with your HDFS. At a minimum, a warning-level message.

C0urante:
[Branch builder tests](https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/994/) succeeded with `--repeat 5` for a custom branch with these changes overlaid onto `5.0.x`.

PR #359, opened 2018-08-07
CC-2493: Improved initialization and logging of Parquet writer

rhauch:
Prior to this change, it was possible that the creation of one Parquet writer failed, and because the HDFS connector’s `TopicPartitionWriter` attempts to retry the call, subsequent calls result in a NPE. This commit always attempts to create the `AvroParquetWriter` when needed (even on subsequent retries), and adds more logging statements to help diagnose the behavior.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/parquet/ParquetRecordWriterProvider.java
+++ b/src/main/java/io/confluent/connect/hdfs/parquet/ParquetRecordWriterProvider.java
@@ -80,8 +80,11 @@ public class ParquetRecordWriterProvider
             );
             log.debug("Opened record writer for: {}", filename);
           } catch (IOException e) {
-            log.debug(
-                "Error creating AvroParquetWriter for file '{}', {}, and schema {}: ",
+            // Ultimately caught and logged in TopicPartitionWriter,
+            // but log in debug to provide more context
+            log.warn(
+                "Error creating {} for file '{}', {}, and schema {}: ",
+                AvroParquetWriter.class.getSimpleName(),
                 filename,
                 compressionCodecName,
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/hdfs/parquet/ParquetRecordWriterProvider.java
```diff
@@ -74,7 +78,15 @@ public void write(SinkRecord record) {
                 true,
                 conf.getHadoopConfiguration()
             );
+            log.debug("Opened record writer for: {}", filename);
           } catch (IOException e) {
+            log.debug(
```
kkonstantine: sounds like a `warn` level here.

src/main/java/io/confluent/connect/hdfs/parquet/ParquetRecordWriterProvider.java
```diff
@@ -74,7 +78,15 @@ public void write(SinkRecord record) {
...
+                "Error creating AvroParquetWriter for file '{}', {}, and schema {}: ",
```
kkonstantine: Would it make sense to use `AvroParquetWriter.class.getSimpleName()` here in case this changes in the future and the log message is left behind?

rhauch:
@kkonstantine, I think I've addressed one of your comments. The other is I think not a concern (see my comment above re WARN level), but since your approval assumes addressing it I'd appreciate your thoughts.

kkonstantine:
I agree, there's some redundancy as in other places, but I feel `warn` makes sense from within a catch block. Not feeling too strong about it either. LGTM

PR #362, opened 2018-08-08
CC-2399: Fix NPE during close

rayokota:
This is to fix NPE from system test:

```
2018-07-19 07:49:57,466 ERROR WorkerSinkTask{id=hdfs-sink-0}
Task threw an uncaught and unrecoverable exception (org.apache.kafka.connect.runtime.WorkerTask)
java.lang.NullPointerException
at io.confluent.connect.hdfs.HdfsSinkTask.close(HdfsSinkTask.java:135)
at org.apache.kafka.connect.runtime.WorkerSinkTask.commitOffsets(WorkerSinkTask.java:376)
at org.apache.kafka.connect.runtime.WorkerSinkTask.closePartitions(WorkerSinkTask.java:560)
at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:183)
at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:170)
at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:214)
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
at java.util.concurrent.FutureTask.run(FutureTask.java:262)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
at java.lang.Thread.run(Thread.java:745)
```

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -118,8 +118,6 @@ public class HdfsSinkTask extends SinkTask {
   public void close(Collection<TopicPartition> partitions) {
     if (hdfsWriter != null) {
       hdfsWriter.close();
-    } else {
-      log.warn("Close called on a null HDFS writer.");
     }
   }
 
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -116,7 +116,9 @@ public class HdfsSinkTask extends SinkTask {
 
   @Override
   public void close(Collection<TopicPartition> partitions) {
-    hdfsWriter.close();
+    if (hdfsWriter != null) {
+      hdfsWriter.close();
+    }
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -118,6 +118,8 @@ public class HdfsSinkTask extends SinkTask {
   public void close(Collection<TopicPartition> partitions) {
     if (hdfsWriter != null) {
       hdfsWriter.close();
+    } else {
+      log.warn("Close called on a null HDFS writer.");
     }
   }
 


kkonstantine:
This appears not and then, but basically means that the connector is asked to closed before it gets even started (see the check for `null` writer in exception handling during start). Do we have an idea of what the route cause of this NPE might be?

rayokota:
@kkonstantine , In the following code:

```
public void execute() {
        initializeAndStart();
        try {
            while (!isStopping())
                iteration();
        } finally {
            // Make sure any uncommitted data has been committed and the task has
            // a chance to clean up its state
            closePartitions();
        }
    }
```

The HDFS connector fails to start in `initializeAndStart` (no Exception is thrown by `HdfsSinkTask.start`.  It then performs `closePartitions`, which calls `close` on an uninitialized `hdfsWriter`

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -116,7 +116,9 @@ public void open(Collection<TopicPartition> partitions) {
 
   @Override
   public void close(Collection<TopicPartition> partitions) {
-    hdfsWriter.close();
+    if (hdfsWriter != null) {
+      hdfsWriter.close();
```
kkonstantine: I think we need a warn message here to say that `close` was called on a `null` `hdfsWriter`.

kkonstantine:
@rayokota exactly. That's a case I was referring to, and usually that's how the NPE appears. Although we've skipped fixing this in the past, NPE is not great here for any reason. I suggest fixing and printing a `WARN` instead. Still, we'd like to know more about why the connector `initializeAndStart` fails without an exception but also without proper initialization (in `start` at least)

rayokota:
@kkonstantine , the `HdfsSinkTask.start` has

```
      log.info("Couldn't start HdfsSinkConnector:", e);
      log.info("Shutting down HdfsSinkConnector.");
      if (hdfsWriter != null) {
        hdfsWriter.close();
        hdfsWriter.stop();
      }
    }
```

The `ConnectException` in this case is `ConnectionRefused` from HDFS.  The exception is swallowed and `info()` messages are logged.

Are you suggesting changing those `info()` to `warn()` as well?

kkonstantine:
I see. Log above didn't contain this info message. I know see that we don't log anything in stop either. Curious this didn't happen more often. I'm fine either way (message or no message) here.

PR #363, opened 2018-08-09
CC-2515: Open temp Parquet files with OVERWRITE mode

rhauch:
Change the `ParquetRecordWriterProvider` class to open the `AvroParquetWriter` instance for each Parquet temp file with ‘overwrite’ mode. This required using the `AvroParquetWriter.Builder` class rather than the deprecated constructor, which did not have a way to specify the mode.

The Avro format's writer already uses an "overwrite" mode, so this brings the same behavior to the Parquet format.

Note that the connector "commits" the records written to a temp file by closing it and moving it to the final location, as part of the HDFS connector's write ahead log behavior. The change in this PR is useful when the connector restarts and there's already an existing Parquet temp file; without this change a `FileAlreadyExistsException` is thrown.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/parquet/ParquetRecordWriterProvider.java
+++ b/src/main/java/io/confluent/connect/hdfs/parquet/ParquetRecordWriterProvider.java
@@ -20,6 +20,7 @@ import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.sink.SinkRecord;
 import org.apache.parquet.avro.AvroParquetWriter;
+import org.apache.parquet.hadoop.ParquetFileWriter;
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/hdfs/parquet/ParquetRecordWriterProvider.java
```diff
@@ -69,15 +70,15 @@ public void write(SinkRecord record) {
...
-                path,
-                avroSchema,
-                compressionCodecName,
-                blockSize,
-                pageSize,
-                true,
-                conf.getHadoopConfiguration()
-            );
+            writer = AvroParquetWriter.<GenericRecord>builder(path)
+                                      .withSchema(avroSchema)
```
kkonstantine: I believe the format in multiline chained calls is still 2 tabs in in connectors. Not a big deal though.

PR #367, opened 2018-08-27
Skip config setting for append-late-data

cyrusv:
S3 and GCS support this feature, but HDFS does not yet

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -379,6 +379,7 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
     Set<String> skip = new HashSet<>();
     skip.add(STORAGE_CLASS_CONFIG);
     skip.add(FORMAT_CLASS_CONFIG);
+    skip.add(APPEND_LATE_DATA);
 
     // Order added is important, so that group order is maintained
     ConfigDef visible = new ConfigDef();


src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -331,11 +331,11 @@ public void alterSchema(String database, String tableName, Schema schema) {
         topicPartitionWriters.put(tp, topicPartitionWriter);
       }
     } catch (
-        ClassNotFoundException
-            | IllegalAccessException
-            | InstantiationException
-            | InvocationTargetException
-            | NoSuchMethodException e
+    ClassNotFoundException
```
kkonstantine: noise and incorrect indentation by this change.

checkstyle/suppressions.xml
```diff
@@ -39,7 +39,7 @@
 
     <suppress
             checks="NPathComplexity"
-            files="(TopicPartitionWriter).java"
+            files="(TopicPartitionWriter|WALFile).java"
```
kkonstantine: Not obvious why this change is required here.

cricket007:
Linking to #256

cyrusv:
@cricket007 Thanks for the link!

I left some notes on the other PR, but it seems like "@asdf2014" isn't responding. To summarize, I think the connector does not need the suggested change, since the expected behavior is to rotate only after the certain amount of time has passed, not any time a record comes that does not fit in the current "window". The connector should handle late records intended for a previous window by opening another writer in the `writers` map.

It does raise a good question, maybe @kkonstantine can shed some light on why S3 connector does rotate if partitions do not match, while HDFS does not have this additional condition? Introduced in `kafka-connect-storage-cloud` commit `60eb3b0fc21cc1431e581c9e9dab1f98ea8`.

kkonstantine:
Don't have memory of the differ you mention @cyrusv at the moment. 
But let's not block on this. 
LGTM

PR #378, opened 2018-10-05
CC-2797:  Avoid using stale value for now

rayokota:
A stale value may be preventing scheduled rotations.


rayokota:
One hypothesis is that `now` is not getting updated because `buffer` is never empty.  Is that possible?

PR #408, opened 2019-01-23
Support for multiple schema types on a single topic

p-pekala:
I would like to handle multiple schema types while keeping compatibility and hive integration turned on. Kafka connect supports it by 'value.subject.name.strategy' property. HDFS sink should be able to write records to different files basing on schema name and create separate hive tables for them.


p-pekala:
[clabot:check]

PR #414, opened 2019-02-15
Add "the License" ref

maxzheng:


rhauch:
The checkstyle report was due to this change moving the constructor out of the range of the suppression rule. Will fix in a separate PR.

maxzheng:
@rhauch Sorry about that and thanks for fixing it!

PR #416, opened 2019-02-26
MINOR: Correct checkstyle suppression line number for constructor

rhauch:
Caused by adding an extra line in #414, which shifted the constructor line number by one and making the suppression rule miss the constructor.

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -45,7 +45,7 @@
     <suppress
             checks="ParameterNumber"
             files="(TopicPartitionWriter).java"
-            lines="112,148"
+            lines="149"
     />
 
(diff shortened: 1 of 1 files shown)


checkstyle/suppressions.xml
```diff
@@ -45,7 +45,7 @@
     <suppress
             checks="ParameterNumber"
             files="(TopicPartitionWriter).java"
-            lines="112,148"
+            lines="149"
```
rhauch: Only one constructor has more than 10 parameters, so the "112" is no longer needed.

rhauch:
I thought about removing lines, but would prefer to not suppress the # of parameters check more than necessary.

PR #417, opened 2019-02-27
CC-3950: Change the HDFS sink task to return to Connect offsets for records committed to HDFS

rhauch:
Modified the task’s `preCommit(…)` method to always return the offsets for only those records that have been committed to HDFS. Connect will then use the consumer to commit those offsets to the `__consumer_offsets` topic, which allows for simple monitoring of the approximate lag of the connector.

First, this change does not affect the HDFS connector’s exactly once behavior, because the connector still writes and uses the offsets it records in the files in HDFS; as long as these files are found in HDFS for the topic partitions, the HDFS connector will use these to determine where Connect should begin consuming. If the HDFS connector does not find the files in HDFS, it will let Connect and the consumer determine where to start, which will be based upon the committed offsets if found or upon the `consumer.auto.offset.reset` value specified in the worker config (which for Connect defaults to `earliest`) if no offsets are found. None of this behavior has changed.

What has changed is that, starting with CP 4.0.1, the HDFS sink connector never returned offsets, which means Connect never updating the offsets in `__consumer_offsets` for the topic partitions the connector read. Users could still change these offsets, and the connector would use these if no offsets were found in HDFS, but it never updated these offsets.

Now, the HDFS connector does return the offsets of the records from the various topic partitions that were most recently committed to HDFS, and Connect will thus periodically (based upon `consumer.offset.flush.interval.ms` in the worker config) write these offsets to the `__consumer_offsets` topic using the appropriate consumer group for the connector.

Because the offsets are committed to `__consumer_offsets` asynchronously, it is likely that the offsets in `__consumer_offsets` will be somewhat behind where the connector actually is. In fact, the offsets in `__consumer_offsets` should never be ahead of the records that the connector has written to HDFS. This allows the offsets as a simple way to estimate the upper limit of the actual connector lag.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -504,7 +504,10 @@ public class DataWriter {
 
   public Map<TopicPartition, Long> getCommittedOffsets() {
     for (TopicPartition tp : assignment) {
-      offsets.put(tp, topicPartitionWriters.get(tp).offset());
+      Long committedOffset = topicPartitionWriters.get(tp).committedOffset();
+      if (committedOffset != null) {
+        offsets.put(tp, committedOffset);
+      }
     }
     return offsets;
   }
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -503,8 +503,10 @@ public class DataWriter {
   }
 
   public Map<TopicPartition, Long> getCommittedOffsets() {
+    log.debug("Writer looking for last offsets for topic partitions {}", assignment);
     for (TopicPartition tp : assignment) {
       Long committedOffset = topicPartitionWriters.get(tp).committedOffset();
+      log.debug("Writer found last offset {} for topic partition {}", committedOffset, tp);
       if (committedOffset != null) {
         offsets.put(tp, committedOffset);
       }
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -25,7 +25,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Collection;
-import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
@@ -119,13 +119,20 @@ public class HdfsSinkTask extends SinkTask {
   public Map<TopicPartition, OffsetAndMetadata> preCommit(
       Map<TopicPartition, OffsetAndMetadata> currentOffsets
   ) {
-    // return an empty map so Connect doesn't commit offsets
-    return Collections.emptyMap();
+    // Although the connector manages offsets via files in HDFS, we still want to have Connect
+    // commit the consumer offsets for records this task has consumed from its topic partitions and
+    // committed to HDFS.
+    Map<TopicPartition, OffsetAndMetadata> result = new HashMap<>();
+    for (Map.Entry<TopicPartition, Long> entry : hdfsWriter.getCommittedOffsets().entrySet()) {
+      result.put(entry.getKey(), new OffsetAndMetadata(entry.getValue()));
+    }
+    log.debug("Returning committed offsets {}", result);
+    return result;
   }
 
   @Override
   public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
-    // Do nothing as the connector manages the offset
+    // Do nothing as the connector manages the offset and overrides 'preCommit()'
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -95,7 +95,8 @@ public class HdfsSinkTask extends SinkTask {
       }
     }
 
-    log.info("HDFS connector does not commit consumer offsets to Kafka. Upon startup, HDFS "
+    log.info("The connector relies on offsets in HDFS filenames, but does commit these offsets to "
+        + "Connect to enable monitoring progress of the HDFS connector. Upon startup, the HDFS "
         + "Connector restores offsets from filenames in HDFS. In the absence of files in HDFS, "
         + "the connector will attempt to find offsets for its consumer group in the "
         + "'__consumer_offsets' topic. If offsets are not found, the consumer will "
@@ -124,15 +125,15 @@ public class HdfsSinkTask extends SinkTask {
     // committed to HDFS.
     Map<TopicPartition, OffsetAndMetadata> result = new HashMap<>();
     for (Map.Entry<TopicPartition, Long> entry : hdfsWriter.getCommittedOffsets().entrySet()) {
+      log.debug(
+          "Found last committed offset {} for topic partition {}",
+          entry.getValue(),
+          entry.getKey()
+      );
       result.put(entry.getKey(), new OffsetAndMetadata(entry.getValue()));
     }
     log.debug("Returning committed offsets {}", result);
     return result;
-  }
-
-  @Override
-  public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
-    // Do nothing as the connector manages the offset and overrides 'preCommit()'
   }
 
   @Override
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -98,6 +98,7 @@ public class TopicPartitionWriter {
   private final AvroData avroData;
   private final Set<String> appended;
   private long offset;
+  private Long committedOffset;
   private final Map<String, Long> startOffsets;
   private final Map<String, Long> offsets;
   private final long timeoutMs;
@@ -208,6 +209,7 @@ public class TopicPartitionWriter {
     state = State.RECOVERY_STARTED;
     failureTime = -1L;
     offset = -1L;
+    committedOffset = null;
     if (writerProvider != null) {
       extension = writerProvider.getExtension();
     } else if (newWriterProvider != null) {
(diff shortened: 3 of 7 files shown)


src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -480,8 +482,13 @@ public void buffer(SinkRecord sinkRecord) {
...
 
-  public long offset() {
-    return offset;
+  /**
+   * Get the offset that was last committed to HDFS, or null if nothing was committed yet.
+   *
+   * @return the last committed offset; may be null
+   */
+  public Long committedOffset() {
+    return committedOffset;
```
rhauch: The `offset` value was *almost* what we needed, but unfortunately it is modified with the first record the writer processes. This means that it does not always represent the last offset that was committed.

Rather than change the logic that relied upon this behavior, I chose to introduce a new `committedOffset` member to represent the last offset that was committed, and set it equal to `offset` whenever the writer commits its file.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -558,6 +565,7 @@ private void readOffset() throws ConnectException {
     );
     if (fileStatusWithMaxOffset != null) {
       offset = FileUtils.extractOffset(fileStatusWithMaxOffset.getPath().getName()) + 1;
+      committedOffset = offset;
```
rhauch: If we read the offset from HDFS files, we update the committed offset here.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -505,7 +505,10 @@ public Partitioner getPartitioner() {
 
   public Map<TopicPartition, Long> getCommittedOffsets() {
     for (TopicPartition tp : assignment) {
-      offsets.put(tp, topicPartitionWriters.get(tp).offset());
+      Long committedOffset = topicPartitionWriters.get(tp).committedOffset();
+      if (committedOffset != null) {
+        offsets.put(tp, committedOffset);
+      }
```
rhauch: We only include the offset for topic partitions with records that were committed by the TopicPartitionWriter instances to HDFS. If `committedOffset()` returns null, we don't even include it.

Despite the name, the previous incarnation of this method didn't return only the committed offsets; as mentioned in another comment, the offset of the first record each TopicPartitionWriter processed would be returned, whether or not this was actually written. And because this method was only used in tests, I felt it was acceptable to change the behavior to return only the committed offset.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
wicknicks: nit: the committedOffset is actually `offset - 1`. `offset` here is the value we will next read from.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -767,6 +775,7 @@ private void commitFile(String encodedPartition) {
     startOffsets.remove(encodedPartition);
     offsets.remove(encodedPartition);
     offset = offset + recordCounter;
+    committedOffset = offset;
```
wicknicks: same here? `committedOffset` is `offset - 1`.

rhauch: I was thinking the same thing - `endOffset` seems to be more accurate.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
kkonstantine: (pressed resolved by accident, apologies). 
Setting 
`committedOffset = FileUtils.extractOffset(fileStatusWithMaxOffset.getPath().getName());`
and then
`offset = committedOffset + 1`
probably makes more sense (fyi `preCommit` won't accidentally commit offsets that have not been consumed, we might delay offset commit with a wrong setting).

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
kkonstantine: I think the choice here is, do we want to report the most accurate number or the most safest number. (`offset-1` vs `endOffset`. I think that's what it is here, but I might have to look closer). Maybe go with the safest option of `endOffset`?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
rhauch: I just made this same change re setting `committedOffset` and `offset`. Great minds... :-D

What do you mean by "we might delay offset commit with a wrong setting"?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
rhauch: Interesting. I tried changing it locally, and it felt awkward (e.g., in the JavaDoc for `offsetToCommit()` because, for the most part, this `TopicPartitionWriter` uses "commit" specifically for committing to HDFS. `offsetToCommit()` just felt like it was the offset that had yet to be written to HDFS.

So I guess I prefer leaving it `committedOffset`.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -119,13 +119,19 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
   public Map<TopicPartition, OffsetAndMetadata> preCommit(
       Map<TopicPartition, OffsetAndMetadata> currentOffsets
   ) {
-    // return an empty map so Connect doesn't commit offsets
-    return Collections.emptyMap();
+    // Although the connector manages offsets via the WAL, we still want to have Connect commit
```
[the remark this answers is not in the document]
rhauch: ack. will reword.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -119,13 +119,19 @@ public void put(Collection<SinkRecord> records) throws ConnectException {
...
+    // the consumer offsets for records this task has consumed from its topic partitions and
+    // committed to HDFS.
+    Map<TopicPartition, OffsetAndMetadata> result = new HashMap<>();
```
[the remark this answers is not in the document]
rhauch: I tried this, but not sure what I was doing wrong. I agree, go streams or go home.

rhauch:
@wicknicks , @kkonstantine: thanks for reviewing. I've added a commit that incorporates most of your suggestions (all but the renaming of `committedOffsets()` to `offsetsToCommit()` as described in the inline comments.

Would appreciate another review.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
kkonstantine: I meant that there's no risk to commit offsets ahead of the current position with `preCommit`. 
See: https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L410-L421

It's just that the recommendation will be ignored and therefore the lag won't be represented as soon as it would be possible.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
kkonstantine: Not insisting here. Just wanted to make an observation. `Commit` is a very overloaded term in these connectors indeed.

src/test/java/io/confluent/connect/hdfs/avro/DataWriterAvroTest.java
```diff
@@ -396,7 +396,7 @@ public void testFlushPartialFile() throws Exception {
     Map<TopicPartition, Long> committedOffsets = hdfsWriter.getCommittedOffsets();
     assertTrue(committedOffsets.containsKey(TOPIC_PARTITION));
     long previousOffset = committedOffsets.get(TOPIC_PARTITION);
-    assertEquals(NUMBER_OF_RECORDS, previousOffset);
+    assertEquals(NUMBER_OF_RECORDS-1, previousOffset);
```
kkonstantine: nit: that tells me we don't run checkstyle on tests (?)

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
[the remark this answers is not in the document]
wicknicks: I just realized, if we plan to backport this to `4.0.x`, then we can't use java 8 apis (unless we want to have a separate PR for `4.0.x`).

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
rhauch: Gotcha. The WorkerSinkTask will prevent committing offsets for records that haven't yet been consumed, but IMO it'd be better to not rely upon that feature since it may result in a larger lag (as measured by the committed offset), especially if the connector always returns a yet-to-be-consumed offset.

I think we're all on the same page, and the updated code uses the correct offset anyway.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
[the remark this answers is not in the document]
rhauch: I'll submit the PR against `4.0.x` and change to imperative style, and create a separate PR that will switch to the functional style (which IMO is far better).

rhauch:
@kkonstantine , @wicknicks: I've rebased the PR on `4.0.x` and switched from functional to imperative code for `HdfsSinkTask.preCommit()`. Otherwise, this is the same set of code as previously reviewed. I've removed the "WIP" from the title and description, so this is ready for the final review.

src/test/java/io/confluent/connect/hdfs/avro/DataWriterAvroTest.java
kkonstantine: @rhauch this needs spaces around `-`

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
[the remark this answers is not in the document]
wicknicks: Makes backporting on these lines a bit harder, otherwise should be alright.

src/test/java/io/confluent/connect/hdfs/avro/DataWriterAvroTest.java
rhauch: fixed

rhauch:
@wicknicks, @kkonstantine: I made a small change to some of the log messages and the formatting of a test. I'll give you a chance to review and disapprove. :-D

PR #422, opened 2019-03-22
CC-4124:  Clear partition assignment on close

rayokota:
When the HDFS Sink task is closed, it does not clear its partition assignment (something that the S3 Sink task does). This can cause an NPE since during `close` it clears the `topicPartitionWriters` but not the `assignment` that is used to iterate over `topicPartitionWriters`.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -458,10 +458,9 @@ public class DataWriter {
         topicPartitionWriters.get(tp).close();
       } catch (ConnectException e) {
         log.error("Error closing writer for {}. Error: {}", tp, e.getMessage());
-      } finally {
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -462,6 +462,7 @@ public void close() {
         topicPartitionWriters.remove(tp);
       }
     }
+    assignment.clear();
```
avocader: Can it be null?

rayokota:
Thanks for the review @avocader ! I made one more small change to match the same exact code in the S3 Sink connector.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
rayokota: No, we iterate over `assignment` right before this statement.

PR #425, opened 2019-04-06
CC-4318 Fix off-by-one error for offset reporting to Kafka topic

cyrusv:
Consumer groups were reporting current-offset as one less than
log-end-offset in the fully caught-up state.

This fixes the topic's perception of current offset to match
the current offset we are storing in HDFS

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -487,7 +487,8 @@ public class TopicPartitionWriter {
    * This method returns the next offset after the last one in HDFS, useful for some APIs
    * (like Kafka Consumer offset tracking).
    *
-   * @return Next offset after the last offset written to HDFS
+   * @return Next offset after the last offset written to HDFS, or -1 if no file has been committed
+   * yet
    */
   public long offset() {
     return offset;
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -488,7 +488,7 @@ public class TopicPartitionWriter {
    * (like Kafka Consumer offset tracking).
    *
    * @return Next offset after the last offset written to HDFS, or -1 if no file has been committed
-   * yet
+   *     yet
    */
   public long offset() {
     return offset;
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -333,8 +333,7 @@ public class TopicPartitionWriter {
             nextState();
           case WRITE_PARTITION_PAUSED:
             if (currentSchema == null) {
-              if (compatibility != StorageSchemaCompatibility.NONE
-                  && offset != -1) {
+              if (compatibility != StorageSchemaCompatibility.NONE && offset != -1) {
                 String topicDir = FileUtils.topicDirectory(url, topicsDir, tp.topic());
                 CommittedFileFilter filter = new TopicPartitionCommittedFileFilter(tp);
                 FileStatus fileStatusWithMaxOffset = FileUtils.fileStatusWithMaxOffset(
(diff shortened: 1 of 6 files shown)


src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -482,13 +482,8 @@ public void buffer(SinkRecord sinkRecord) {
...
   }
 
-  /**
-   * Get the offset that was last committed to HDFS, or null if nothing was committed yet.
-   *
-   * @return the last committed offset; may be null
-   */
-  public Long committedOffset() {
-    return committedOffset;
+  public long offset() {
```
rhauch: @cyrusv, I'm not sure I understand why we're returning 1 past what was last committed to HDFS. We need to verify that what is committed to the consumer offsets exactly matches what is committed to HDFS. I'm not convinced this really identified/fixed the bug.

Second, `offset` is initialized to `-1L`, so is it ever possible that this method return that? Before this change, the offset getter might return null, but the code in DataWriter looked for this case and never wrote that into the offsets.

Finally, we lost JavaDoc - can we put it back to explain what offset this actually returns?

rhauch: So I wonder if this is the source of the confusion: the consumer offset should point to the offset of the *next record to be consumed, whereas the prior change recorded for the consumer offset the *last record committed to HDFS*.

It'd be good to document this, though I still think the `offset()` method should mention why it's 1 past the last committed offset and why (to match the consumer offsets). Also, we still need to handle the -1 case mentioned above.

cyrusv: Thanks for the feedback, @rhauch  -- I've renamed some variables and added some documentation to clarify the work done here. You can see the work is most closely related to `284fbc2`, for reference.

cyrusv: I decided to stick to a primitive with `-1` initial value, since a lot of logic in the `TopicPartitionWriter` tests against `-1` and I think keeping this code as-is (relative to pre-`284fbc2`) with a clearer name will provide the most stability. I could be convinced that `null` initial value would be preferable if we'd rather have NPEs than misleading long values, even though I decided it wasn't worth the tradeoff since we've seen this structure work successfully in many production envs.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -483,12 +483,13 @@ public void buffer(SinkRecord sinkRecord) {
   }
 
   /**
-   * Get the offset that was last committed to HDFS, or null if nothing was committed yet.
-   *
-   * @return the last committed offset; may be null
+   * HDFS Connector tracks offsets in filenames in HDFS (for Exactly Once Semantics) as the last
+   * record's offset that was written to the last file in HDFS.
+   * This method returns the next offset after the last one in HDFS, useful for some APIs
+   * (like Kafka Consumer offset tracking).
```
rhauch: This is not a well-formed JavaDoc (missing `@return`), and it should also mention when this method return `-1`.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -483,12 +483,13 @@ public void buffer(SinkRecord sinkRecord) {
...
    */
-  public Long committedOffset() {
-    return committedOffset;
+  public long committedConsumerOffset() {
```
rhauch: Again, this method doesn't know about the consumer -- this offset literally is 1 past the offset of the last record committed to HDFS. Perhaps rename it to `nextOffset()` instead (as suggested by your test).

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -776,8 +778,7 @@ private void commitFile(String encodedPartition) {
     storage.commit(tempFile, committedFile);
     startOffsets.remove(encodedPartition);
     offsets.remove(encodedPartition);
-    offset = offset + recordCounter; // offset of next record to be reading
-    committedOffset = offset - 1; // offset of last record written to HDFS
+    committedConsumerOffset = committedConsumerOffset + recordCounter;
```
rhauch: If we keep `offset` and remove `committedConsumerOffset`, then we could keep the old line 779 (with it's comment) and remove the old 780.

PR #437, opened 2019-06-10
Logical type source task

liukrimhrim:
Updated the source connector task, so the source connector can produce both primitive type schema and logical type schema for system tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,9 +63,16 @@
         <repository>
             <id>confluent</id>
             <name>Confluent</name>
-            <url>${confluent.maven.repo}</url>
+            <url>http://packages.confluent.io/maven/</url>
         </repository>
     </repositories>
+    <pluginRepositories>
+        <pluginRepository>
+            <id>confluent</id>
+            <name>Confluent</name>
+            <url>http://packages.confluent.io/maven/</url>
+        </pluginRepository>
+    </pluginRepositories>
 
     <dependencies>
         <dependency>
--- a/src/main/java/io/confluent/connect/hdfs/tools/SchemaSourceConnector.java
+++ b/src/main/java/io/confluent/connect/hdfs/tools/SchemaSourceConnector.java
@@ -35,7 +35,9 @@ public class SchemaSourceConnector extends SourceConnector {
 
   @Override
   public void start(Map<String, String> props) {
+
     this.config = props;
+    System.out.println("fuck source");
   }
 
   @Override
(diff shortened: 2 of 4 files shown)


src/main/java/io/confluent/connect/hdfs/tools/SchemaSourceConnector.java
```diff
@@ -35,6 +35,7 @@ public String version() {
 
   @Override
   public void start(Map<String, String> props) {
+
```
kkonstantine: nit: white noise

src/main/java/io/confluent/connect/hdfs/tools/SchemaSourceTask.java
```diff
@@ -41,6 +47,9 @@
   private static final Logger log = LoggerFactory.getLogger(SchemaSourceTask.class);
   private static final String ID_FIELD = "id";
   private static final String SEQNO_FIELD = "seqno";
+  private static final String SCHEMA_TYPE = "schema.type";
+  private static final String CONNECT_AVRO_DECIMAL_PRECISION_PROP = "connect.decimal.precision";
```
kkonstantine: Where is this config originally declared? Can we import the member variable directly here instead of just the string value?

src/main/java/io/confluent/connect/hdfs/tools/SchemaSourceTask.java
```diff
@@ -62,6 +71,25 @@
       .field("id", Schema.INT32_SCHEMA)
       .field("seqno", Schema.INT64_SCHEMA)
       .build();
+
+
+  private static Schema valueSchemaLogical = SchemaBuilder.struct().version(1).name("record")
```
kkonstantine: let's give those variables names other than foobar1, foobar2. What's the difference between them?

src/main/java/io/confluent/connect/hdfs/tools/SchemaSourceTask.java
```diff
@@ -127,6 +158,89 @@ public void start(Map<String, String> props) {
...
+    if ("logical".equalsIgnoreCase(this.schemaType)) {
+      return pollLogical();
+    } else if ("primitive".equalsIgnoreCase(this.schemaType)) {
+      return pollPrimitive();
+    } else {
+      throw new InterruptedException("Invalid schema type");
+    }
+  }
+
+  private List<SourceRecord> pollLogical() throws InterruptedException {
```
kkonstantine: This actually could extend the schema that is created by poll primitive. But even without that, I wonder if those methods could share some code.

pom.xml
```diff
@@ -63,9 +63,16 @@
         <repository>
             <id>confluent</id>
             <name>Confluent</name>
-            <url>${confluent.maven.repo}</url>
+            <url>http://packages.confluent.io/maven/</url>
         </repository>
     </repositories>
+    <pluginRepositories>
```
kkonstantine: Any idea why these changes are needed in the pom?

PR #446, opened 2019-07-29
CC-5739: fix bug connector retries nonretriable ConnectionException

liukrimhrim:
This bug fix addresses https://confluentinc.atlassian.net/browse/CC-5739 and  https://confluentinc.atlassian.net/browse/ESCALATION-1447

Before the bug fix, HDFS connector retries nonretriable `ConnectionException` indefinitely.

After the fix, HDFS connector will fail the task without any retry if `ConnectionException` is thrown, just like S3 connector does


liukrimhrim:
My fix is based off the master branch. The 3 failures in `FailureRecoveryTest` were there in the master already, not caused by my code.

liukrimhrim:
Not a good approach

PR #451, opened 2019-08-24
CC-6089: Improve fault tolerance of HDFS sink connector

rhauch:
Changed the task to throw an exception if the task could not be started successfully. In some cases, an exception would be logged but not re-thrown.

Also checked for null references in a couple of places that can result from an improperly-started or failing task.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -67,7 +67,7 @@ public class DataWriter {
   private static final Time SYSTEM_TIME = new SystemTime();
   private final Time time;
 
-  private Map<TopicPartition, TopicPartitionWriter> topicPartitionWriters;
+  private final Map<TopicPartition, TopicPartitionWriter> topicPartitionWriters;
   private String url;
   private HdfsStorage storage;
   private String topicsDir;
@@ -451,11 +451,20 @@ public class DataWriter {
     // data may have continued to be processed and we'd have to restart from the recovery stage,
     // make sure we apply the WAL, and only reuse the temp file if the starting offset is still
     // valid. For now, we prefer the simpler solution that may result in a bit of wasted effort.
-    for (TopicPartition tp : assignment) {
+    for (Map.Entry<TopicPartition, TopicPartitionWriter> ent : topicPartitionWriters.entrySet()) {
       try {
-        topicPartitionWriters.get(tp).close();
+        TopicPartitionWriter writer = ent.getValue();
+        if (writer != null) {
+          // In some failure modes, with might not have created a writer for all assignments
+          writer.close();
+        }
       } catch (ConnectException e) {
-        log.error("Error closing writer for {}. Error: {}", tp, e.getMessage());
+        log.warn(
+            "Unable to close writer for topic partition {}: {}",
+            ent.getKey(),
+            e.getMessage(),
+            e
+        );
       }
     }
     topicPartitionWriters.clear();
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -455,7 +455,7 @@ public class DataWriter {
       try {
         TopicPartitionWriter writer = ent.getValue();
         if (writer != null) {
-          // In some failure modes, with might not have created a writer for all assignments
+          // In some failure modes, the writer might not have been created for all assignments
           writer.close();
         }
       } catch (ConnectException e) {
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -451,20 +451,14 @@ public class DataWriter {
     // data may have continued to be processed and we'd have to restart from the recovery stage,
     // make sure we apply the WAL, and only reuse the temp file if the starting offset is still
     // valid. For now, we prefer the simpler solution that may result in a bit of wasted effort.
-    for (Map.Entry<TopicPartition, TopicPartitionWriter> ent : topicPartitionWriters.entrySet()) {
+    for (TopicPartitionWriter writer : topicPartitionWriters.values()) {
       try {
-        TopicPartitionWriter writer = ent.getValue();
         if (writer != null) {
           // In some failure modes, the writer might not have been created for all assignments
           writer.close();
         }
       } catch (ConnectException e) {
-        log.warn(
-            "Unable to close writer for topic partition {}: {}",
-            ent.getKey(),
-            e.getMessage(),
-            e
-        );
+        log.warn("Unable to close writer for topic partition {}: ", writer.topicPartition(), e);
       }
     }
     topicPartitionWriters.clear();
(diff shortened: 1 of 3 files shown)


src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -87,12 +87,24 @@ public void start(Map<String, String> props) {
...
+          } finally {
+            log.debug("Stopping data writer due to task start failure.");
+            hdfsWriter.stop();
+          }
+        } catch (Throwable t) {
+          log.debug("Error closing and stopping data writer: {}", t.getMessage(), t);
+        }
       }
+      // Always throw the original exception that prevent us from starting
+      throw e;
```
rhauch: It seems really strange that the task caught but did not rethrow the ConnectException. (NPEs and other runtimetimes would not be caught and would be handled by Connect.) By catching but not rethrowing the exception, Connect does not know task startup failed and will continue using the task.

This makes me wonder if this was intentional, though the existing log message still says "Shutting down HdfsSinkConnector".

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -451,11 +451,20 @@ public void close() {
...
+          writer.close();
+        }
       } catch (ConnectException e) {
-        log.error("Error closing writer for {}. Error: {}", tp, e.getMessage());
+        log.warn(
+            "Unable to close writer for topic partition {}: {}",
+            ent.getKey(),
+            e.getMessage(),
+            e
+        );
```
rhauch: It's possible that the DataWriter is instantiated, the assignments set, but the TopicPartitionWriters are not created (again, due to an error). In this case, the close method attempts to fine the TPW for each assignment, and because there are none we get an NPE.

The new code iterates over any existing TPW we've created and closes them safely, ensuring to close all of them and log any errors. Because this is close, IMO we can log and don't need to propagate the exceptions.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -692,15 +692,15 @@ private void writeRecord(SinkRecord record) {
   }
 
   private void closeTempFile(String encodedPartition) {
-    if (writers.containsKey(encodedPartition)) {
-      io.confluent.connect.storage.format.RecordWriter writer = writers.get(encodedPartition);
+    io.confluent.connect.storage.format.RecordWriter writer = writers.remove(encodedPartition);
+    if (writer != null) {
```
rhauch: This is a slightly more efficient approach, and more tolerant should there not be a writer for an encoded partition.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -451,11 +451,20 @@ public void close() {
...
+        if (writer != null) {
+          // In some failure modes, with might not have created a writer for all assignments
+          writer.close();
+        }
       } catch (ConnectException e) {
-        log.error("Error closing writer for {}. Error: {}", tp, e.getMessage());
+        log.warn(
+            "Unable to close writer for topic partition {}: {}",
+            ent.getKey(),
+            e.getMessage(),
```
wicknicks: nit: this will print the message twice.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -451,11 +451,20 @@ public void close() {
     // data may have continued to be processed and we'd have to restart from the recovery stage,
     // make sure we apply the WAL, and only reuse the temp file if the starting offset is still
     // valid. For now, we prefer the simpler solution that may result in a bit of wasted effort.
-    for (TopicPartition tp : assignment) {
+    for (Map.Entry<TopicPartition, TopicPartitionWriter> ent : topicPartitionWriters.entrySet()) {
       try {
-        topicPartitionWriters.get(tp).close();
+        TopicPartitionWriter writer = ent.getValue();
+        if (writer != null) {
+          // In some failure modes, with might not have created a writer for all assignments
```
wicknicks:
```suggestion
          // In some failure modes, the writer might not have been created for all assignments
```

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -451,11 +451,20 @@ public void close() {
     // data may have continued to be processed and we'd have to restart from the recovery stage,
     // make sure we apply the WAL, and only reuse the temp file if the starting offset is still
     // valid. For now, we prefer the simpler solution that may result in a bit of wasted effort.
-    for (TopicPartition tp : assignment) {
+    for (Map.Entry<TopicPartition, TopicPartitionWriter> ent : topicPartitionWriters.entrySet()) {
```
wicknicks: nit: should we `topicPartitionWriters.values()` here instead?

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
wicknicks: great catch! I wonder if `ConnectException` is throw by any of the code below? this pattern of catching this exception and swallowing it seems to appear in multiple places. maybe that's why this went unnoticed?.

rhauch: It definitely happens: DataWriter constructor throws ConnectException with any problems.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
rhauch: Yeah, that makes sense. Had to add a getter in `TopicPartitionWriter` to get the (final) `TopicPartition`, but that's easy and worth the simplicity here.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
rhauch: ack.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
rhauch: Something occurred to me: by removing the writer *before* attempting to close it, we will lose track of the (un)closed writer should there be an exception during `close()` on the next few lines.

In fact, I think if we don't do this, then if an exception happens during `close()`, the existing code would not remove the `RecordWriter`, which means that `getWriter(String encodedPartition)` would return this improperly closed writer, and the `TopicPartitionWriter` would continue to use the writer even though it is an unexpected state. This is in contrast to the contract for `Closeable`, which is that once `close()` is called the resource not be used anymore. I think we see evidence of this: [a particular NPE in the ParquetWriter](https://issues.apache.org/jira/browse/PARQUET-544) is often attributed to `close()` being called more than once, and the `TopicPartitionWriter` would do this if the first `close()` failed and the `RecordWriter` were left in this `writers` map (as is the case in the current code).

So, I think that we should remove the `RecordWriter` regardless of whether or not we're able to successfully `close()` the writer -- as long as we at least call `close()`, which the current PR does. The other benefit is that when `close()` does fail and the task ultimately retries the topic partition, the `getWriter(...) method will not see in the `writers` map the writer for which `close()` failed and will create a new one and put it into the `writers` map. This is the behavior we want, as subsequent attempts to retry writing records to the topic partition will be performed with a writer re-acquired after the `close()` failure and hopefully when HDFS has healed itself.

I did add another commit that documents this reasoning in a comment just before this line of code.

PR #456, opened 2019-09-16
CC-6196: fix NPE for store.url config

levzem:
`hfds.url` was deprecated everywhere in the documentation for `store.url` but it was not reflected in the code and results in NPE

addresses the issue in [DOCS-1449](https://confluentinc.atlassian.net/browse/DOCS-1449)
and [here](https://confluentinc.atlassian.net/browse/CC-6196) 

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -14,6 +14,7 @@
 
 package io.confluent.connect.hdfs;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
@@ -302,11 +303,11 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
     partitionerConfig = new PartitionerConfig(partitionerConfigDef, originalsStrings());
     this.name = parseName(originalsStrings());
     this.hadoopConfig = new Configuration();
-    this.url = extractUrl(props);
     addToGlobal(hiveConfig);
     addToGlobal(partitionerConfig);
     addToGlobal(commonConfig);
     addToGlobal(this);
+    this.url = extractUrl();
   }
 
   public static Map<String, String> addDefaults(Map<String, String> props) {
@@ -336,22 +337,21 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
    * Returns the url property. Preference is given to property <code>store.url</code> over
    * <code>hdfs.url</code> because <code>hdfs.url</code> is deprecated.
    *
-   * @param props - Map of properties for the connector
-   * @return String - url for HDFS
+   * @return String url for HDFS
    */
-  private String extractUrl(Map<String, String> props) {
-    String storageUrl = props.get(StorageCommonConfig.STORE_URL_CONFIG);
-    if (storageUrl != null && !storageUrl.equals(StorageCommonConfig.STORE_URL_DEFAULT)) {
+  private String extractUrl() {
+    String storageUrl = getString(StorageCommonConfig.STORE_URL_CONFIG);
+    if (StringUtils.isNotBlank(storageUrl)) {
       return storageUrl;
     }
 
-    String hdfsUrl = props.get(HDFS_URL_CONFIG);
-    if (hdfsUrl != null && !hdfsUrl.equals(HDFS_URL_DEFAULT)) {
+    String hdfsUrl = getString(HDFS_URL_CONFIG);
+    if (StringUtils.isNotBlank(hdfsUrl)) {
       return hdfsUrl;
     }
 
     throw new ConfigException(
-        String.format("Configuration %s cannot be empty."), StorageCommonConfig.STORE_URL_CONFIG
+        String.format("Configuration %s cannot be empty.", StorageCommonConfig.STORE_URL_CONFIG)
     );
   }
 
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -332,7 +332,14 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
     }
   }
 
-  private static String extractUrl(Map<String, String> props) {
+  /**
+   * Returns the url property. Preference is given to property `store.url` over `hdfs.url` because
+   * `hdfs.url` is deprecated.
+   *
+   * @param props - Map of properties for the connector
+   * @return String - url for HDFS
+   */
+  private String extractUrl(Map<String, String> props) {
     String storageUrl = props.get(StorageCommonConfig.STORE_URL_CONFIG);
     if (storageUrl != null && !storageUrl.equals(StorageCommonConfig.STORE_URL_DEFAULT)) {
       return storageUrl;
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -302,7 +302,7 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
     partitionerConfig = new PartitionerConfig(partitionerConfigDef, originalsStrings());
     this.name = parseName(originalsStrings());
     this.hadoopConfig = new Configuration();
-    this.url = extractUrl(originalsStrings());
+    this.url = extractUrl(props);
     addToGlobal(hiveConfig);
     addToGlobal(partitionerConfig);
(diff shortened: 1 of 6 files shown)


levzem:
[clabot:check]

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -301,6 +302,7 @@ protected HdfsSinkConnectorConfig(ConfigDef configDef, Map<String, String> props
     partitionerConfig = new PartitionerConfig(partitionerConfigDef, originalsStrings());
     this.name = parseName(originalsStrings());
     this.hadoopConfig = new Configuration();
+    this.url = extractUrl(originalsStrings());
```
wicknicks: why do we need to make a copy here? just use props directly?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -330,10 +332,30 @@ private void addConfig(Map<String, ?> parsedProps, ComposableConfig config) {
     }
   }
 
+  private static String extractUrl(Map<String, String> props) {
+    String storageUrl = props.get(StorageCommonConfig.STORE_URL_CONFIG);
+    if (storageUrl != null && !storageUrl.equals(StorageCommonConfig.STORE_URL_DEFAULT)) {
+      return storageUrl;
+    }
+
+    String hdfsUrl = props.get(HDFS_URL_CONFIG);
```
wicknicks: HDFS url should be prioritized above the store url?

levzem: since HDFS is being deprecated I figured that we should prefer `store.url`

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: following the general pattern as above but I will change

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
wicknicks: I would mention this in the javadoc.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -330,10 +332,30 @@ private void addConfig(Map<String, ?> parsedProps, ComposableConfig config) {
     }
   }
 
+  private static String extractUrl(Map<String, String> props) {
```
wicknicks: nit: this method doesn't have to be static.

src/test/java/io/confluent/connect/hdfs/HdfsSinkTaskWithSecureHDFSTest.java
```diff
@@ -54,6 +55,8 @@ public void testSinkTaskPut() throws Exception {
         sinkRecords.add(sinkRecord);
       }
     }
+
+    properties.remove(StorageCommonConfig.STORE_URL_CONFIG);
```
wicknicks: wondering if we should set the value of STORE_URL_CONFIG prop in io.confluent.connect.hdfs.TestWithSecureMiniDFSCluster#createProps (that is where the url for HDFS_URL_CONFIG is being overridden).

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -330,10 +332,37 @@ private void addConfig(Map<String, ?> parsedProps, ComposableConfig config) {
     }
   }
 
+  /**
+   * Returns the url property. Preference is given to property `store.url` over `hdfs.url` because
+   * `hdfs.url` is deprecated.
```
wicknicks: backticks don't really format anything here.

src/test/java/io/confluent/connect/hdfs/HdfsSinkTaskWithSecureHDFSTest.java
```diff
@@ -14,6 +14,7 @@
 
 package io.confluent.connect.hdfs;
 
+import io.confluent.connect.storage.common.StorageCommonConfig;
```
wicknicks: not used anymore?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -330,10 +332,37 @@ private void addConfig(Map<String, ?> parsedProps, ComposableConfig config) {
...
+      return storageUrl;
+    }
+
+    String hdfsUrl = props.get(HDFS_URL_CONFIG);
+    if (hdfsUrl != null && !hdfsUrl.equals(HDFS_URL_DEFAULT)) {
+      return hdfsUrl;
+    }
+
+    throw new ConfigException(
+        String.format("Configuration %s cannot be empty."), StorageCommonConfig.STORE_URL_CONFIG
```
wicknicks: nit: do you think we should say `"Configuration STORE_URL_CONFIG and HDFS_URL_DEFAULT should not be empty"` instead?

levzem: if it falls to this case, both are empty and since `hdfs.url` is deprecated it makes sense to encourage the use of `store.url`

kkonstantine: your call to `format` has a bug. And the fact that we don't see it tells me we don't have good unit tests. 
`HdfsSinkConnectorConfigTest` is the place to add some.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -330,10 +332,37 @@ private void addConfig(Map<String, ?> parsedProps, ComposableConfig config) {
...
+  /**
+   * Returns the url property. Preference is given to property <code>store.url</code> over
+   * <code>hdfs.url</code> because <code>hdfs.url</code> is deprecated.
+   *
+   * @param props - Map of properties for the connector
+   * @return String - url for HDFS
+   */
+  private String extractUrl(Map<String, String> props) {
+    String storageUrl = props.get(StorageCommonConfig.STORE_URL_CONFIG);
+    if (storageUrl != null && !storageUrl.equals(StorageCommonConfig.STORE_URL_DEFAULT)) {
```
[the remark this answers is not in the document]
kkonstantine: But also taking a step back, not sure what are we aiming for here.

For example, not sure why we don't use `getString`
There's no very elegant way to write this. Here's a suggestion that minimizes lines (instead of calls): 
```
    if (StringUtils.isNotBlank(getString(STORE_URL_CONFIG))) {
      return getString(STORE_URL_CONFIG);
    }
    if (StringUtils.isNotBlank(getString(HDFS_URL_CONFIG))) {
      return getString(HDFS_URL_CONFIG);
    }
    throw new ConfigException(
        String.format("Configuration %s cannot be empty.", StorageCommonConfig.STORE_URL_CONFIG
    );
```

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -330,10 +332,37 @@ private void addConfig(Map<String, ?> parsedProps, ComposableConfig config) {
...
+        String.format("Configuration %s cannot be empty."), StorageCommonConfig.STORE_URL_CONFIG
+    );
+  }
+
   public String getName() {
     return name;
   }
 
+  public String getUrl() {
+    return url;
```
kkonstantine: Given that this is called in `DataWriter` anyways, I believe we can just keep the logic in here and not introduce `extractUrl`

levzem: considering it gets called multiple places I think it's fine to preprocess

kkonstantine: It's good to keep in mind that it's not in a critical path and `url` is not used heavily. So reducing boiler place and number of declared methods could also be seen as an optimization. Not a blocker though

src/test/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfigTest.java
```diff
@@ -50,6 +51,19 @@ public void setUp() throws Exception {
...
 
+  @Test(expected = ConfigException.class)
+  public void testUrlConfigMustBeNonEmpty() {
+    properties.remove(StorageCommonConfig.STORE_URL_CONFIG);
+    properties.remove(HdfsSinkConnectorConfig.HDFS_URL_CONFIG);
+    connectorConfig = new HdfsSinkConnectorConfig(properties);
+  }
+
+  @Test
+  public void testStorageCommonUrlPreferred() {
```
kkonstantine: May we also have a test were store is undefined but hdfs is?

PR #460, opened 2019-10-04
CC-6423: Disable FileSystem Object Caching

gharris1727:
When two tasks run on the same worker, they grab an aliased FileSystem object from a static cache. If one task shuts down early (e.g. due to failing to acquire the lease on the WAL), they will close the shared FileSystem object. The other task holding that FIleSystem object will then throw exceptions when trying to use that object, causing other tasks on the same worker to crash.

There are two ways to solve this, either leave the shared FileSystem object open (resource leak), or create a new instance whose lifetime is controlled by the task (disabling caching). This PR implements the latter strategy, and all call sites of `FileSystem.get` are replaced with `FileSystem.newInstance`, and the lifetime is managed manually.

This PR also includes a test to verify that closing the shared FileSystem object does not impact the functionality of the task.

This supersedes #426


cyrusv:
@gharris1727 , Great work diagnosing the problem, and I think some patterns here are really nice. I'm confused by the inconsistency between String, Json, and Avro Record Writer Providers -- I see in the String case, you've cleaned up the giant try-catch, but in Json you haven't. 

I also see in Avro case you're not using Storage.create() at all. I realize this would involve a change in `AvroFormat` to pass the storage object through, which may have felt out-of-scope for the problem, but unless I missed something it's a simple fix that can keep our classes better aligned. This will help keep the abstraction consistent that the FS object is tied to an HDFS Storage instance, which makes intuitive sense, without exceptionally doing a weird dance in the Avro case.

I need to re-read and think harder on the WAL case since it's a little different, and I'm not sure I understand the testing for it.

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -316,6 +314,7 @@ public synchronized void close() throws IOException {
         // Close the underlying stream iff we own it...
         if (ownOutputStream) {
           out.close();
+          fs.close();
```
rhauch: If `fs` is null if we don't own the output stream, wouldn't it be safer to have a finally block and then call `fs.close()` when `fs != null`?

We also need to handle exceptions properly, as mentioned in an earlier comment.

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -462,10 +461,8 @@ public Reader(Configuration conf, Option... opts) throws IOException {
         initialize(filename, file, start, len, conf, headerOnly != null);
       } catch (RemoteException re) {
         log.error("Failed creating a WAL Reader: " + re.getMessage());
-        if (re.getClassName().equals(WALConstants.LEASE_EXCEPTION_CLASS_NAME)) {
-          if (fs != null) {
-            fs.close();
-          }
+        if (fs != null) {
+          fs.close();
```
rhauch: We need to handle exceptions properly, as mentioned in an earlier comment. Here `fs.close()` can fail, yet we want to still throw `re`.

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -652,6 +649,7 @@ public synchronized void close() throws IOException {
 
       // Close the input-stream
       in.close();
+      fs.close();
```
rhauch: We need to handle exceptions properly here, as mentioned in an earlier comment.

gharris1727:
@rhauch thanks for calling out those resource leaks and swallowed exceptions. Please take another look to confirm the fixes are adequate.

src/main/java/io/confluent/connect/hdfs/avro/AvroRecordWriterProvider.java
```diff
@@ -85,6 +85,12 @@ public void write(SinkRecord record) {
           }
         } catch (IOException e) {
           throw new DataException(e);
+        } finally {
+          try {
+            close();
+          } catch (Throwable t) {
+            log.error("Could not close Record Writer", t);
```
rhauch: You may want to verify in a scratch area that the DataException will still be thrown. I do recall that an exception on the finally block will override the one caught before the finally block, but I don't recall what happens if the finally block catches it.

src/main/java/io/confluent/connect/hdfs/string/StringRecordWriterProvider.java
```diff
@@ -70,6 +70,12 @@ public void write(SinkRecord record) {
           writer.newLine();
         } catch (IOException e) {
           throw new ConnectException(e);
+        } finally {
+          try {
+            close();
+          } catch (Throwable t) {
+            log.error("Could not close Record Writer", t);
+          }
```
rhauch: This isn't correct here, because we're always calling `close()` after every call to `write(...)`. I was asking about the case when we catch the exception on 72, and possibly closing it here. But do we know whether we're always closing the record writer? If so, we don't need this here. (Please add a comment describing what you find.)

src/main/java/io/confluent/connect/hdfs/wal/WALFile.java
```diff
@@ -308,17 +312,27 @@ private void writeFileHeader()
...
+        keySerializer.close();
+        valSerializer.close();
+        if (out != null) {
+          // Close the underlying stream iff we own it...
+          if (ownOutputStream) {
+            out.close();
+          } else {
+            out.flush();
+          }
+          out = null;
```
rhauch: Should this be in the try-finally block? Right now, if `out.close()` or `out.flush()` throws an exception, we'll not null `out` and, if `close()` is called more than once (albeit unnecessarily) then we'd call close or flush another time. It's probably simpler to just ensure we null `out` in the finally block.

src/main/java/io/confluent/connect/hdfs/avro/AvroRecordWriterProvider.java
gharris1727: I've verified that the original exception gets thrown correctly.

src/main/java/io/confluent/connect/hdfs/string/StringRecordWriterProvider.java
gharris1727: You're right that this is incorrect, I got a little trigger-happy with the finally blocks. It appears that exceptions thrown from this method get caught in [TopicParititionWriter::write](https://github.com/confluentinc/kafka-connect-hdfs/blob/e13532024920e1433f741119427517583adf8e5e/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L409-L414).

The RecordWriter is closed in these cases:
* `HdfsSinkTask::start()` when an exception occurs while recovering the WAL
* `TopicParititonWriter::write()` when in state `SHOULD_ROTATE`
* `TopicPartitionWriter::write()` after a successful write
* `HdfsSinkTask::close()` when partitions are reassigned
* `HdfsSinkTask::stop()` when the task is stopped entirely

I definitely don't understand the retry strategy of the HDFS connector completely, but it appears that the error is being swallowed, so the framework does not trigger the sink record to be retried. If the write failed, all of the previous records that were in the process of being written are potentially lost as well, and need to be retried from the beginning. I can only assume that the connector already handles that properly, otherwise we would be seeing large data loss events whenever HDFS hiccups.

I think i'd err towards leaving the code as-is (as in rolling back my changes entirely), and not closing the writer or it's associated filesystem. They appear to be adequately cleaned up during task shutdown and partition reassignment, and I assume that the existing error handling hits those code-paths.

src/main/java/io/confluent/connect/hdfs/storage/HdfsStorage.java
```diff
@@ -88,14 +92,53 @@ public boolean create(String filename) {
...
   }
 
   @Override
   public OutputStream create(String filename, HdfsSinkConnectorConfig conf, boolean overwrite) {
+    final Path path = new Path(filename);
     try {
-      Path path = new Path(filename);
-      return path.getFileSystem(conf.getHadoopConfiguration()).create(path);
+      return new OutputStream() {
+        FileSystem fs = FileSystem.newInstance(path.toUri(), conf.getHadoopConfiguration());
```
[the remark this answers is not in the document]
savulchik: @wicknicks It seems like `FileSystem#newInstance(URI, Configuration)` puts a unique `FileSystem` object in the `CACHE` (don't know why) using `FileSystem.Cache#getUnique`. So I guess it is safe to assume that you actually get a new fs instance.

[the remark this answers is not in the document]
gharris1727: This call in particular passes in an `HdfsSinkConnectorConfig`, which overrides the `HdfsSinkConnectorConfig` which was used to construct the current instance of `HdfsStorage`. If these two configurations are different, then re-using the same FileSystem does not provide the correct semantics for this method call.

I don't believe this method is called in the HDFS connector, but if it is, we support this method call and provide the correct semantics.

rhauch:
@gharris1727 if you've not already, please run a system test branch builder (running the HDFS system tests) with this PR, and add a link to the build.

gharris1727:
System test build passed: https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/2575/

PR #475, opened 2019-11-28
Handle NullPointerException on missing partition writers

NikolayTsvetkov:
While doing DataWriter.write() method calls it's possible to end up with NPE in case the TopicPartitionWrites is missing for a given TopicPartition. That NPE results in losing the running tasks on the worker.
This commit improves the exception handling and the write() method structure.


gharris1727:
This has been resolved by #496

PR #477, opened 2019-12-13
CC-7247: allow topics.dir config to be more configurable

levzem:
see CC ticket and relevant one pager for explanation

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -72,7 +72,7 @@ public class DataWriter {
   private final Map<TopicPartition, TopicPartitionWriter> topicPartitionWriters;
   private String url;
   private HdfsStorage storage;
-  private String topicsDir;
+  private HashMap<String, String> topicDirs;
   private Format format;
   private RecordWriterProvider writerProvider;
   private io.confluent.connect.storage.format.RecordWriterProvider<HdfsSinkConnectorConfig>
@@ -207,7 +207,6 @@ public class DataWriter {
       }
 
       url = connectorConfig.getUrl();
-      topicsDir = connectorConfig.getString(StorageCommonConfig.TOPICS_DIR_CONFIG);
 
       @SuppressWarnings("unchecked")
       Class<? extends HdfsStorage> storageClass = (Class<? extends HdfsStorage>) connectorConfig
@@ -219,8 +218,16 @@ public class DataWriter {
           url
       );
 
-      createDir(topicsDir);
-      createDir(topicsDir + HdfsSinkConnectorConstants.TEMPFILE_DIRECTORY);
+      topicDirs = new HashMap<>();
+      for (TopicPartition tp : context.assignment()) {
+        topicDirs.computeIfAbsent(tp.topic(), topic -> connectorConfig.getTopicDirFromTopic(topic));
+      }
+
+      for (String directories : topicDirs.values()) {
+        createDir(directories);
+        createDir(directories + HdfsSinkConnectorConstants.TEMPFILE_DIRECTORY);
+      }
+
       String logsDir = connectorConfig.getString(HdfsSinkConnectorConfig.LOGS_DIR_CONFIG);
       createDir(logsDir);
 
@@ -388,7 +395,7 @@ public class DataWriter {
 
     try {
       for (String topic : topics) {
-        String topicDir = FileUtils.topicDirectory(url, topicsDir, topic);
+        String topicDir = FileUtils.topicDirectory(url, topicDirs.get(topic), topic);
         CommittedFileFilter filter = new TopicCommittedFileFilter(topic);
         FileStatus fileStatusWithMaxOffset = FileUtils.fileStatusWithMaxOffset(
             storage,
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -72,6 +72,7 @@ public class DataWriter {
   private final Map<TopicPartition, TopicPartitionWriter> topicPartitionWriters;
   private String url;
   private HdfsStorage storage;
+  private HashMap<String, String> logDirs;
   private HashMap<String, String> topicDirs;
   private Format format;
   private RecordWriterProvider writerProvider;
@@ -220,16 +221,22 @@ public class DataWriter {
 
       topicDirs = new HashMap<>();
       for (TopicPartition tp : context.assignment()) {
-        topicDirs.computeIfAbsent(tp.topic(), topic -> connectorConfig.getTopicDirFromTopic(topic));
+        topicDirs.computeIfAbsent(tp.topic(), top -> connectorConfig.getTopicsDirFromTopic(top));
       }
 
-      for (String directories : topicDirs.values()) {
-        createDir(directories);
-        createDir(directories + HdfsSinkConnectorConstants.TEMPFILE_DIRECTORY);
+      for (String directory : topicDirs.values()) {
+        createDir(directory);
+        createDir(directory + HdfsSinkConnectorConstants.TEMPFILE_DIRECTORY);
       }
 
-      String logsDir = connectorConfig.getString(HdfsSinkConnectorConfig.LOGS_DIR_CONFIG);
-      createDir(logsDir);
+      logDirs = new HashMap<>();
+      for (TopicPartition tp : context.assignment()) {
+        logDirs.computeIfAbsent(tp.topic(), topic -> connectorConfig.getLogsDirFromTopic(topic));
+      }
+
+      for (String directory : logDirs.values()) {
+        createDir(directory);
+      }
 
       // Try to instantiate as a new-style storage-common type class, then fall back to old-style
       // with no parameters
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -94,12 +94,13 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
   // Storage group
   public static final String TOPIC_REGEX_CAPTURE_GROUP_CONFIG = "topic.regex.capture.group";
   public static final String TOPIC_REGEX_CAPTURE_GROUP_DISPLAY = "Topic Regex Capture Group";
-  public static final String TOPIC_REGEX_CAPTURE_GROUP_DOC = "A regex that specifies what groups "
-      + "to capture in the topic, so when specifying `${1}` in `topics.dir`, `${1}` will refer to "
-      + "the first captured group. Example config value of `[a-zA-Z]*` will capture all characters "
-      + "between [-._0-9], so for `topic.dir = ${1}/${2}` and `topic = topic.name_suffix` "
-      + "the corresponding `topic.dir` will be `topic/name/`. By default, this functionality is "
-      + "not enabled.";
+  public static final String TOPIC_REGEX_CAPTURE_GROUP_DOC = "A regex that matches the entire "
+      + "topic and  specifies what groups to capture in the topic, so when specifying `${1}` in "
+      + "`topics.dir`, `${1}` will refer to the first captured group. Example config value of "
+      + "`([a-zA-Z]*)_([a-zA-Z]*)` match a topic that is two words delimited by a - and will "
+      + "capture both words as separate groupds, so for `topic.dir = ${1}/${2}` and "
+      + "`topic = topic_name` the corresponding `topic.dir` will be `topic/name/`. By default, "
+      + "this functionality is not enabled.";
   public static final String TOPIC_REGEX_CAPTURE_GROUP_DEFAULT = null;
   
   // Security group
@@ -437,33 +438,38 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
     if (topixRegexCaptureGroup != null) {
 
       // find all of the captured groups by the regex
-      List<String> topicGroups = new ArrayList<>();
       Matcher matcher = topixRegexCaptureGroup.matcher(topic);
-      while (matcher.find()) {
-        if (!matcher.group().isEmpty()) {
-          topicGroups.add(matcher.group());
-        }
+      if (!matcher.matches()) {
+        throw new ConfigException(
+            String.format(
+                "Configuration %s with value %s does not fully match the specified regex %s in %s",
+                TOPICS_DIR_CONFIG,
+                topicsDir,
+                topixRegexCaptureGroup.pattern(),
+                TOPIC_REGEX_CAPTURE_GROUP_CONFIG
+            )
+        );
       }
 
       // make sure that all references to captured groups actually exist
       if (!partsToReplace.isEmpty()) {
         int largestVar = partsToReplace.get(partsToReplace.size() - 1);
-        if (largestVar > topicGroups.size()) {
+        if (largestVar > matcher.groupCount()) {
           throw new ConfigException(
               String.format(
-                  "Topic %s must be split into at least %d parts using regex pattern %s, "
-                      + "but was actually split into %d parts",
+                  "Topic %s must have at least %d capture groups using regex pattern %s, "
+                      + "but actually had %d capture groups.",
                   topic,
                   largestVar,
                   topixRegexCaptureGroup.pattern(),
-                  topicGroups.size()
+                  matcher.groupCount()
               )
           );
         }
       }
 
-      for (int index : partsToReplace) {
-        topicsDir = topicsDir.replace("${" + index + "}", topicGroups.get(index - 1));
+      for (int i = 1; i < matcher.groupCount() + 1; i++) {
+        topicsDir = topicsDir.replace("${" + i + "}", matcher.group(i));
       }
     }
 
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -16,6 +16,7 @@
 package io.confluent.connect.hdfs;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
@@ -58,6 +59,9 @@ import static io.confluent.connect.storage.common.StorageCommonConfig.STORAGE_CL
 import static io.confluent.connect.storage.common.StorageCommonConfig.STORAGE_CLASS_DISPLAY;
 import static io.confluent.connect.storage.common.StorageCommonConfig.STORAGE_CLASS_DOC;
 import static io.confluent.connect.storage.common.StorageCommonConfig.TOPICS_DIR_CONFIG;
+import static io.confluent.connect.storage.common.StorageCommonConfig.TOPICS_DIR_DEFAULT;
+import static io.confluent.connect.storage.common.StorageCommonConfig.TOPICS_DIR_DISPLAY;
+import static io.confluent.connect.storage.common.StorageCommonConfig.TOPICS_DIR_DOC;
 
 public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
 
@@ -93,7 +97,7 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
   public static final String TOPIC_REGEX_CAPTURE_GROUP_DOC = "A regex that specifies what groups "
       + "to capture in the topic, so when specifying `${1}` in `topics.dir`, `${1}` will refer to "
       + "the first captured group. Example config value of `[a-zA-Z]*` will capture all characters "
-      + "delimited by . _  and -, so for `topic.dir = ${1}/${2}` and `topic = topic.name.suffix` "
+      + "between [-._0-9], so for `topic.dir = ${1}/${2}` and `topic = topic.name_suffix` "
       + "the corresponding `topic.dir` will be `topic/name/`. By default, this functionality is "
       + "not enabled.";
   public static final String TOPIC_REGEX_CAPTURE_GROUP_DEFAULT = null;
@@ -332,6 +336,7 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
   private final Map<String, ComposableConfig> propertyToConfig = new HashMap<>();
   private final Set<AbstractConfig> allConfigs = new HashSet<>();
   private Configuration hadoopConfig;
+  private List<Integer> partsToReplace;
 
   public HdfsSinkConnectorConfig(Map<String, String> props) {
     this(newConfigDef() , addDefaults(props));
@@ -361,6 +366,8 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
           e
       );
     }
+
+    partsToReplace = getPartsToReplace();
   }
 
   public static Map<String, String> addDefaults(Map<String, String> props) {
@@ -439,22 +446,24 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
       }
 
       // make sure that all references to captured groups actually exist
-      int maxVar = extractMaximumInt(topicsDir);
-      if (maxVar > topicGroups.size()) {
-        throw new ConfigException(
-            String.format(
-                "Topic %s must be split into at least %d parts using regex pattern %s, "
-                    + "but was actually split into %d parts",
-                topic,
-                maxVar,
-                topixRegexCaptureGroup.pattern(),
-                topicGroups.size()
-            )
-        );
+      if (!partsToReplace.isEmpty()) {
+        int largestVar = partsToReplace.get(partsToReplace.size() - 1);
+        if (largestVar > topicGroups.size()) {
+          throw new ConfigException(
+              String.format(
+                  "Topic %s must be split into at least %d parts using regex pattern %s, "
+                      + "but was actually split into %d parts",
+                  topic,
+                  largestVar,
+                  topixRegexCaptureGroup.pattern(),
+                  topicGroups.size()
+              )
+          );
+        }
       }
 
-      for (int i = 0; i < maxVar; i++) {
-        topicsDir = topicsDir.replace("${" + (i + 1) + "}", topicGroups.get(i));
+      for (int index : partsToReplace) {
+        topicsDir = topicsDir.replace("${" + index + "}", topicGroups.get(index - 1));
       }
     }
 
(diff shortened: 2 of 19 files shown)


src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -458,6 +550,22 @@ public static ConfigDef getConfig() {
         FORMAT_CLASS_RECOMMENDER
     );
 
+    visible.define(
+        "topics.dir",
+        Type.STRING,
+        "topics",
+        Importance.HIGH,
+        "Top level directory to store the data ingested from Kafka. "
```
ncliang: Make these constants?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -369,6 +415,51 @@ public String getUrl() {
...
+      List<String> topicGroups = new ArrayList<>();
+      Matcher matcher = topixRegexCaptureGroup.matcher(topic);
+      while (matcher.find()) {
+        if (!matcher.group().isEmpty()) {
+          topicGroups.add(matcher.group());
+        }
+      }
+
+      // make sure that all references to captured groups actually exist
+      int maxVar = extractMaximumInt(topicsDir);
```
ncliang: I don't understand this part? I thought topicsDir can contain numbers in them - "/topic_1000" by itself should be a valid topicDir? Or if the topic name contains numbers and ${topic} appears in topicDir, wouldn't your maxVar be not what you expect? I think you can do this with a regex looking for anything that looks like "\$\{\d\}" in your topicsDir and replace with what you find in topicGroups. Catch `IndexOutOfBoundsException` and rethrow the `ConfigException` for referencing non-existing capture group.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -469,6 +577,32 @@ private static void addAllConfigKeys(ConfigDef container, ConfigDef other, Set<S
...
 
+  /**
+   * Extract the biggest integer contained in the string
+   *
+   * @param string - the string to extract the maximum integer from
+   * @return int - the biggest integer found in the string, 0 if none found
+   */
+  private static int extractMaximumInt(String string) {
+    int num = 0;
+    int res = 0;
```
ncliang: Integer overflow? I actually don't think you need this method at all.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: big derp :P

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: thats actually a lot smarter
thanks!

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -92,7 +96,7 @@
   public static final String TOPIC_REGEX_CAPTURE_GROUP_DOC = "A regex that specifies what groups "
       + "to capture in the topic, so when specifying `${1}` in `topics.dir`, `${1}` will refer to "
       + "the first captured group. Example config value of `[a-zA-Z]*` will capture all characters "
-      + "delimited by . _  and -, so for `topic.dir = ${1}/${2}` and `topic = topic.name.suffix` "
+      + "between [-._0-9], so for `topic.dir = ${1}/${2}` and `topic = topic.name_suffix` "
       + "the corresponding `topic.dir` will be `topic/name/`. By default, this functionality is "
```
ncliang: Since the parameter is called `topic`, maybe it would be less confusing to not have the string "topic" appear again in the example? Something like "example_name.suffix" -> "example/name" ?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -551,19 +560,18 @@ public static ConfigDef getConfig() {
...
         Type.STRING,
-        "topics",
+        TOPICS_DIR_DEFAULT,
         Importance.HIGH,
-        "Top level directory to store the data ingested from Kafka. "
-            + "Supports ``${topic}`` in the value, which will be replaced by the actual"
-            + " topic name. Supports ``${1}``, ..., ``${n}`` in conjunction with "
-            + TOPIC_REGEX_CAPTURE_GROUP_CONFIG + ". See " + TOPIC_REGEX_CAPTURE_GROUP_CONFIG
-            + " configuration documentation for details.",
+        TOPICS_DIR_DOC + " Supports ``${topic}`` in the value, which will be "
```
ncliang: Why this doc string is not a constant like the rest?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -578,29 +586,26 @@ private static void addAllConfigKeys(ConfigDef container, ConfigDef other, Set<S
...
+   * @return list of all numbers found in the regex pattern
    */
-  private static int extractMaximumInt(String string) {
-    int num = 0;
-    int res = 0;
+  private List<Integer> getPartsToReplace() {
 
-    for (char c : string.toCharArray()) {
+    List<Integer> toReplace = new ArrayList<>();
+    Pattern pattern = Pattern.compile("\\$\\{\\d+}");
```
ncliang: Do you need to escape the end bracket? Looks weird with mismatched escapes :-/

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -578,29 +586,26 @@ private static void addAllConfigKeys(ConfigDef container, ConfigDef other, Set<S
...
-        num = num * 10 + (c - '0');
-      } else {
-        res = Math.max(res, num);
-
-        // Reset the number
-        num = 0;
+    while (partsMatcher.find()) {
+      String part = partsMatcher.group();
+      if (!part.isEmpty()) {
+        part = part.substring(2, part.length() - 1); // trim "${" and "}"
```
ncliang: You can use match groups for this. With something like "\\$\\{(\d+)\\}" as pattern, match group 1 will be the digits only.

src/test/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfigTest.java
```diff
@@ -84,6 +84,21 @@ public void testTopicDirFromTopicParts() {
     );
   }
 
+  @Test
+  public void testTopicDirCanContainNumber() {
```
ncliang: Add a testcase to test something that is smaller than the range of valid match groups - `${0}`, `${-1000}` or something.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -578,29 +586,26 @@ private static void addAllConfigKeys(ConfigDef container, ConfigDef other, Set<S
...
+        toReplace.add(Integer.valueOf(part));
       }
     }
 
-    return Math.max(res, num);
+    Collections.sort(toReplace);
```
ncliang: No need to sort here just to get the max. Just use `Collections.max()` outside in the caller. Makes it clearer what the code outside is doing, is slightly faster, and doesn't rely on this method returning a sorted list.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -360,6 +365,8 @@ protected HdfsSinkConnectorConfig(ConfigDef configDef, Map<String, String> props
           e
       );
     }
+
+    partsToReplace = getPartsToReplace();
```
ncliang: Why does `partsToReplace` need to be a field?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -438,22 +445,24 @@ public String getTopicDirFromTopic(String topic) {
...
+                  topicGroups.size()
+              )
+          );
+        }
       }
 
-      for (int i = 0; i < maxVar; i++) {
-        topicsDir = topicsDir.replace("${" + (i + 1) + "}", topicGroups.get(i));
+      for (int index : partsToReplace) {
+        topicsDir = topicsDir.replace("${" + index + "}", topicGroups.get(index - 1));
```
ncliang: Another way to do this that might be even nicer is to make `topicGroups` a Map<String, String> mapping "${1}" -> "captured_value_1", "${2}" -> "captured_value_2", etc. Then you can do something like 
```
topicGroups.forEach((k, v) -> topicsDir.replaceAll(k, v));
```
If you still see something that looks like "${\d+}" after that, user tried to reference invalid capture group, throw.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: its redundant to escape the second bracket - i can but i get a warning

src/test/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfigTest.java
levzem: great edge case!

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: micro optimization to not compile the regex pattern and perform the regex search every call, but I just got a better idea - Ill just keep track of the largest index instead

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: I changed the implementation + behavior slightly (see one pager for full explanation) so the code is slightly different not, but let me know what you think

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: fixed

src/test/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfigTest.java
```diff
@@ -53,6 +54,107 @@ public void setUp() throws Exception {
...
+    properties.put(StorageCommonConfig.TOPICS_DIR_CONFIG, topicDir);
+    connectorConfig = new HdfsSinkConnectorConfig(properties);
+
+    connectorConfig.getTopicsDirFromTopic(topic);
+  }
+
+  @Test(expected = ConfigException.class)
+  public void testInvalidTopicDirNegative() {
+    String topic = "a.b.c.d";
+    String topicDir = "${-1}-${2}-${3}-${4}";
```
ncliang: Maybe another case with ${0}. I think that would be the more common mistake.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -399,6 +463,143 @@ public Configuration getHadoopConfiguration() {
...
+   * @param config - String - the config to validate
+   */
+  private void validateReplacements(String config) {
+    // remove all valid ${} substitutions
+    Pattern pattern = Pattern.compile("\\$\\{(\\d+)}");
+    Matcher partsMatcher = pattern.matcher(getString(config));
+    String dir = partsMatcher.replaceAll("").replace("${topic}", "");
+
+    // check for invalid ${} substitutions
+    Pattern invalid = Pattern.compile("\\$\\{.*}");
```
ncliang: nit: these patterns can be made `private static final` and reused.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -75,12 +84,30 @@
...
+  public static final String TOPIC_REGEX_CAPTURE_GROUP_DOC = "A regex that matches the entire "
+      + "topic and  specifies what groups to capture in the topic, so when specifying `${1}` in "
+      + "`topics.dir`, `${1}` will refer to the first captured group. Example config value of "
+      + "`([a-zA-Z]*)_([a-zA-Z]*)` match a topic that is two words delimited by a - and will "
+      + "capture both words as separate groupds, so for `topic.dir = ${1}/${2}` and "
+      + "`topic = example_name` the corresponding `topic.dir` will be `example/name/`. By default, "
+      + "this functionality is not enabled.";
+  public static final String TOPIC_REGEX_CAPTURE_GROUP_DEFAULT = null;
+
+  private static final String DIR_REGEX_DOC = " Supports ``${topic}`` in the value, which will be "
```
gharris1727: `${0}` and `${topic}` would be the same if the regex matched the whole topic , which we could choose to enforce. I guess the cost of adding `${topic}` is pretty low, and this feature could be enabled without ever specifying a capture config like `^.*$`

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -281,9 +308,28 @@ public static ConfigDef newConfigDef() {
         FORMAT_CLASS_RECOMMENDER,
         AVRO_COMPRESSION_RECOMMENDER);
 
+    int lastOrder = 0;
+    String group = "Storage";
     for (ConfigDef.ConfigKey key : storageConfigDef.configKeys().values()) {
       configDef.define(key);
+      group = key.group;
+      lastOrder = key.orderInGroup;
```
gharris1727: Does `.values()` always iterate over these in-order?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -314,6 +363,21 @@ protected HdfsSinkConnectorConfig(ConfigDef configDef, Map<String, String> props
     addToGlobal(commonConfig);
     addToGlobal(this);
     this.url = extractUrl();
+    try {
+      this.topixRegexCaptureGroup = getString(TOPIC_REGEX_CAPTURE_GROUP_CONFIG) != null
+          ? Pattern.compile(getString(TOPIC_REGEX_CAPTURE_GROUP_CONFIG))
```
gharris1727: Pull this out to a variable

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -399,6 +463,143 @@ public Configuration getHadoopConfiguration() {
...
   }
 
+  /**
+   * Performs all substitutions on `logs.dir` and calculates the final directory for a topic
+   *
+   * @param topic - String - the topic whose directory to find
+   * @return String - the directory name and path
+   */
+  public String getLogsDirFromTopic(String topic) {
+    return getDirFromTopic(getString(LOGS_DIR_CONFIG), topic, logDirGroupsMaxIndex);
```
gharris1727: Have you verified that you've replaced all of the call-sites for `getString(LOGS_DIR_CONFIG)` and `getString(TOPICS_DIR_CONFIG)` with your new function?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -399,6 +463,143 @@ public Configuration getHadoopConfiguration() {
...
+  private String getDirFromTopic(String dir, String topic, int maxIndexParts) {
+    dir = dir.replace("${topic}", topic);
+
+    // only if configured
+    if (topixRegexCaptureGroup != null) {
+
+      // find all of the captured groups by the regex
+      Matcher matcher = topixRegexCaptureGroup.matcher(topic);
+      if (!matcher.matches()) {
+        throw new ConfigException(
```
gharris1727: You can provide the configuration and its value to the `ConfigException` for better formatting of the error.
```suggestion
        throw new ConfigException(
            TOPIC_REGEX_CAPTURE_GROUP_CONFIG,
            topixRegexCaptureGroup.pattern(),
```

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -399,6 +463,143 @@ public Configuration getHadoopConfiguration() {
...
+                    + "but actually had %d capture groups.",
+                topic,
+                maxIndexParts,
+                topixRegexCaptureGroup.pattern(),
+                matcher.groupCount()
+            )
+        );
+      }
+
+      for (int index = 1; index < matcher.groupCount() + 1; index++) {
```
gharris1727: Might as well give them all the groups
```suggestion
      for (int index = 0; index < matcher.groupCount() + 1; index++) {
```

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -399,6 +463,143 @@ public Configuration getHadoopConfiguration() {
...
+    Matcher partsMatcher = pattern.matcher(getString(config));
+    String dir = partsMatcher.replaceAll("").replace("${topic}", "");
+
+    // check for invalid ${} substitutions
+    Pattern invalid = Pattern.compile("\\$\\{.*}");
+    Matcher invalidMatcher = invalid.matcher(dir);
+    if (invalidMatcher.find()) {
+      throw new ConfigException(
+          String.format(
+              "%s: %s contains an invalid ${} substitution %s",
```
gharris1727: Can they escape these substitutions like `\$\{topic\}` and have their directory be literally `${topic}/some-topic`? Would this check prevent this use-case? Maybe it's better to be left out.
You can actually generate a list of these substitutions and then shell out to the connect-utils validator. Maybe we can use this if we don't target versions that are included in CP.

src/test/java/io/confluent/connect/hdfs/HdfsSinkConnectorTestBase.java
```diff
@@ -73,6 +72,9 @@
     props.put(PartitionerConfig.LOCALE_CONFIG, "en");
     props.put(PartitionerConfig.TIMEZONE_CONFIG, "America/Los_Angeles");
 
+    props.put(HdfsSinkConnectorConfig.TOPIC_REGEX_CAPTURE_GROUP_CONFIG, "(.*)[\\.\\-](.*)");
```
gharris1727: Do these belong in the test base class?
Maybe we can scope these to just the relevant test classes? idk, i'm kinda 50/50 on this.
It does mean that we're not testing with this config disabled, which probably isn't good.

I think you can keep the refactor for topicsDir as a Map, but just disable the config options.

src/test/java/io/confluent/connect/hdfs/wal/WALFileTest.java
```diff
@@ -36,10 +36,12 @@
   @Test
   public void testAppend() throws Exception {
     setUp();
+    properties.put(HdfsSinkConnectorConfig.TOPIC_REGEX_CAPTURE_GROUP_CONFIG, "(.*)");
```
gharris1727: why is `testAppend` special?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -75,12 +84,30 @@
   private static final String HADOOP_HOME_DOC = "The Hadoop home directory.";
   private static final String HADOOP_HOME_DISPLAY = "Hadoop home directory";
 
+  // Storage group
+  public static final String TOPIC_REGEX_CAPTURE_GROUP_CONFIG = "topic.regex.capture.group";
+  public static final String TOPIC_REGEX_CAPTURE_GROUP_DISPLAY = "Topic Regex Capture Group";
+  public static final String TOPIC_REGEX_CAPTURE_GROUP_DOC = "A regex that matches the entire "
+      + "topic and  specifies what groups to capture in the topic, so when specifying `${1}` in "
+      + "`topics.dir`, `${1}` will refer to the first captured group. Example config value of "
```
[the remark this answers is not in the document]
levzem: I won't be supporting `${0}` but otherwise changed

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -75,12 +84,30 @@
...
+      + "`([a-zA-Z]*)_([a-zA-Z]*)` match a topic that is two words delimited by a - and will "
```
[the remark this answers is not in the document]
levzem: making me feel like a certain someone that needs grammarly

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: I didn't really wanna support `${0}`, `${topic}` is more intuitive I think

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: ‾\_(ツ)_/‾

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: yes

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: `values()` is called on a hashmap so actually order isn't guaranteed

levzem: fixed

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: initially said I wouldn't support but maybe I should

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: was going to say can't use `connect-utils` but depends on what version - this can be a WIP - this check still catches all `${-n}` values if they are somehow silly enough to put negative numbers. honestly why would they escape and if they do... let it break

src/test/java/io/confluent/connect/hdfs/HdfsSinkConnectorTestBase.java
levzem: I put this in the test base class because it is the easiest way to "integration" test this config to make sure it works

src/test/java/io/confluent/connect/hdfs/wal/WALFileTest.java
levzem: it uses a different topic that isn't captured by the normal regex

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -385,7 +399,7 @@ public void syncWithHive() throws ConnectException {
 
     try {
       for (String topic : topics) {
-        String topicDir = FileUtils.topicDirectory(url, topicsDir, topic);
+        String topicDir = FileUtils.topicDirectory(url, topicDirs.get(topic), topic);
```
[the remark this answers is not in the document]
levzem: this satisfies requirements so I am not sure this is a necessary change. We can easily extend the functionality later if it is required

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -458,6 +660,18 @@ public static ConfigDef getConfig() {
...
     );
 
+    visible.define(
+        TOPICS_DIR_CONFIG,
+        Type.STRING,
+        TOPICS_DIR_DEFAULT,
+        Importance.HIGH,
+        TOPICS_DIR_DOC + DIR_REGEX_DOC,
+        "Storage",
+        2,
```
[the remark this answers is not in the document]
levzem: it's the order in `StorageCommonConfig` unfortunately I don't see a better way than hardcoding

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -75,12 +84,30 @@
...
+
+  // HDFS Group
   public static final String LOGS_DIR_CONFIG = "logs.dir";
   public static final String LOGS_DIR_DOC =
-      "Top level directory to store the write ahead logs.";
+      "Top level directory to store the write ahead logs." + DIR_REGEX_DOC;
   public static final String LOGS_DIR_DEFAULT = "logs";
   public static final String LOGS_DIR_DISPLAY = "Logs directory";
-
+
```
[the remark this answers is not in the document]
levzem: I don't see a double line in my source code after `LOGS_DIR_DISPLAY`
github shows double lines sometimes incorrectly

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
levzem: this is only run once at ConfigConstruction essentially being static - but I can make it static

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -292,9 +341,12 @@ public static ConfigDef newConfigDef() {
   private final StorageCommonConfig commonConfig;
   private final HiveConfig hiveConfig;
   private final PartitionerConfig partitionerConfig;
+  private final Pattern topixRegexCaptureGroup;
```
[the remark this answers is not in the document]
levzem: el oh el

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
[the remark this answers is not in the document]
gharris1727: It's not a double line, it's a line that was empty before (`\n\n`) and now it has some spaces on it (`\n   \n`)

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -399,6 +465,141 @@ public Configuration getHadoopConfiguration() {
...
+      if (!matcher.matches()) {
+        throw new ConfigException(
+            TOPIC_REGEX_CAPTURE_GROUP_CONFIG,
+            topixRegexCaptureGroup.pattern(),
+            String.format("Topic %s does not fully match the specified regex.", topic)
+        );
+      }
+
+      // make sure that all references to captured groups actually exist
+      if (maxIndexParts > 0 && maxIndexParts > matcher.groupCount()) {
```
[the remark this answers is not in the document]
levzem: yes

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -399,6 +465,141 @@ public Configuration getHadoopConfiguration() {
...
+
+    return toReplace.isEmpty() ? 0 : Collections.max(toReplace);
+  }
+
+  /**
+   * Validates that the `topics.dir`, `logs.dir`, and `topic.regex.capture.group` configs are all
+   * within valid ranges.
+   */
+  private void validateDirsAndRegex() {
+    if (topicDirGroupsMaxIndex > 0 && topixRegexCaptureGroup == null) {
```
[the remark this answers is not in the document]
levzem: fixed to return -1

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -399,6 +465,141 @@ public Configuration getHadoopConfiguration() {
...
+                    + "but actually had %d capture groups.",
+                topic,
+                maxIndexParts,
+                topixRegexCaptureGroup.pattern(),
+                matcher.groupCount()
+            )
+        );
+      }
+
+      for (int index = 0; index < matcher.groupCount() + 1; index++) {
```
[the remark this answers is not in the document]
levzem: I think is an unnecessary micro optimization

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -399,6 +465,141 @@ public Configuration getHadoopConfiguration() {
...
+  }
+
+  /**
+   * Performs all substitutions and calculates the final directory for a topic
+   *
+   * @param dir - String - the directory to perform substitutions on
+   * @param topic - String - the topic whose directory to find
+   * @return String - the directory name and path
+   */
+  private String getDirFromTopic(String dir, String topic, int maxIndexParts) {
```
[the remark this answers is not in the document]
levzem: it helps with validation instead of recomputing it every time - I don't think we necessarily gain anything by removing the instance variables

PR #294, opened 2018-03-09
Supports ORC Format w/Hive

cricket007:
Closes #173 

Just putting this here for visibility @buom ; Please let us know the state of your code.

PR #484, opened 2020-01-23
CC-7648: add ORC format with Hive

levzem:
taking over from #294 

thank you for your contribution @buom!


src/main/java/io/confluent/connect/hdfs/orc/OrcFileReader.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Iterator;
+
+public class OrcFileReader implements SchemaFileReader<HdfsSinkConnectorConfig, Path> {
+
+  private static final Logger log = LoggerFactory.getLogger(OrcFileReader.class);
+
+  @Override
+  public Schema getSchema(HdfsSinkConnectorConfig conf, Path path) {
+    try {
+      log.info("Opening ORC record reader for: {}", path);
```
gharris1727: How often is this called? I think this should be at most debug severity.

src/main/java/io/confluent/connect/hdfs/orc/OrcFileReader.java
```diff
@@ -0,0 +1,118 @@
...
+              schemaType = Schema.Type.ARRAY;
+              break;
+            case MAP:
+              schemaType = Schema.Type.MAP;
+              break;
+            default:
+              throw new DataException("Unknown type " + fieldObjectInspector.getCategory().name());
+          }
+
+          schemaBuilder.field(schema.getFieldName(), SchemaBuilder.type(schemaType).build());
```
gharris1727: Does this support arbitrarily nested schemas?
Does it support specifying types for `LIST/ARRAY` and `MAP`?

src/test/java/io/confluent/connect/hdfs/orc/OrcDataFileReader.java
```diff
@@ -0,0 +1,49 @@
...
+  @Override
+  public Collection<Object> readData(Configuration conf, Path path) throws IOException {
+    ArrayList<Object> collection = new ArrayList<>();
+    OrcFile.ReaderOptions readerOptions = new OrcFile.ReaderOptions(conf);
+    Reader reader = OrcFile.createReader(path, readerOptions);
+
+    RecordReader rows = reader.rows();
+
+    Object row = null;
+    for (; rows.hasNext(); ) {
```
kkonstantine: aka `while` loop

src/main/java/io/confluent/connect/hdfs/orc/OrcFileReader.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        return schemaBuilder.build();
+      } else {
+        throw new ConnectException(
+            "Top level type must be of type STRUCT, but was "
```
gharris1727: Is this necessary?
Why doesn't it support primitive root types?

src/test/java/io/confluent/connect/hdfs/orc/OrcHiveUtilTest.java
```diff
@@ -0,0 +1,173 @@
...
+import org.apache.hadoop.hive.ql.metadata.Table;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.connect.data.Field;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.apache.kafka.connect.sink.SinkTaskContext;
+import org.junit.Test;
+
+import java.util.*;
```
kkonstantine: we avoid wildcards in imports. Small adjustment to your IDE should fix this. (by default the threshold is quite low).

src/main/java/io/confluent/connect/hdfs/orc/OrcHiveUtil.java
```diff
@@ -0,0 +1,105 @@
...
+    table.setSerializationLib(getHiveOrcSerde());
+
+    try {
+      table.setInputFormatClass(getHiveOrcInputFormat());
+      table.setOutputFormatClass(getHiveOrcOutputFormat());
+    } catch (HiveException e) {
+      throw new HiveMetaStoreException("Cannot find input/output format:", e);
+    }
+
+    // convert copycat schema schema to Hive columns
```
gharris1727: haha is this a stale comment from one of the other HiveUtils?
```suggestion
    // convert connect schema to Hive columns
```

src/main/java/io/confluent/connect/hdfs/orc/OrcHiveUtil.java
```diff
@@ -0,0 +1,105 @@
...
+      String tableName,
+      Schema schema,
+      Partitioner partitioner
+  ) throws HiveMetaStoreException {
+
+    Table table = newTable(database, tableName);
+    table.setTableType(TableType.EXTERNAL_TABLE);
+    table.getParameters().put("EXTERNAL", "TRUE");
+
+    // tableName is always the topic name
```
gharris1727: Is this assumption still true?
Does this assumption affect the behavior of the other Hive Utils?

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
```diff
@@ -0,0 +1,122 @@
...
+  private static Logger log = LoggerFactory.getLogger(OrcRecordWriterProvider.class);
+  private static final String EXTENSION = ".orc";
+
+  @Override
+  public String getExtension() {
+    return EXTENSION;
+  }
+
+  @Override
+  public RecordWriter getRecordWriter(final HdfsSinkConnectorConfig conf, final String filename) {
```
gharris1727: nit
```suggestion
  public RecordWriter getRecordWriter(HdfsSinkConnectorConfig conf, String filename) {
```

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
```diff
@@ -0,0 +1,122 @@
...
+    final Path path = new Path(filename);
```
gharris1727: nit
```suggestion
    Path path = new Path(filename);
```

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+public class OrcRecordWriterProvider implements RecordWriterProvider<HdfsSinkConnectorConfig> {
+
+  private static Logger log = LoggerFactory.getLogger(OrcRecordWriterProvider.class);
```
kkonstantine: nit: can be `final`

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
```diff
@@ -0,0 +1,122 @@
...
+              log.info("Opening ORC record writer for: {}", filename);
+              writer = OrcFile
+                  .createWriter(path, OrcFile.writerOptions(conf.getHadoopConfiguration())
+                      .inspector(objectInspector)
+                      .callback(writerCallback));
+            }
+          }
+
+          if (schema.type() == Schema.Type.STRUCT) {
+            if (log.isTraceEnabled()) {
```
gharris1727: Don't need a guard clause because topic/kafkaPartition/kafkaOffset are cheap calls.
```suggestion
```

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.kafka.connect.sink.SinkRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+public class OrcRecordWriterProvider implements RecordWriterProvider<HdfsSinkConnectorConfig> {
+
+  private static Logger log = LoggerFactory.getLogger(OrcRecordWriterProvider.class);
+  private static final String EXTENSION = ".orc";
```
gharris1727: And maybe this should get pushed to OrcUtil and used in the tests.
```suggestion
  public static final String EXTENSION = ".orc";
```

src/main/java/io/confluent/connect/hdfs/orc/OrcFileReader.java
```diff
@@ -0,0 +1,118 @@
...
+                + reader.getObjectInspector().getCategory().name()
+        );
+      }
+    } catch (IOException e) {
+      throw new ConnectException("Failed to get schema for file " + path, e);
+    }
+  }
+
+  public boolean hasNext() {
```
kkonstantine: Missing `@Override`

kkonstantine: Here and below

src/main/java/io/confluent/connect/hdfs/orc/OrcFileReader.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public Object next() {
+    throw new UnsupportedOperationException();
+  }
+
+  public void remove() {
+    throw new UnsupportedOperationException();
+  }
+
+  public Iterator<Object> iterator() {
```
kkonstantine: Missing `@Override`
and `@Nonnull` (from `javax.annotation.Nonnull`)

src/main/java/io/confluent/connect/hdfs/orc/OrcUtil.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  /**
+   * Convert a Struct into a Writable array
+   */
+  public static Object[] convertStruct(Struct struct) {
+    List<Object> data = new LinkedList<>();
+    for (Field field : struct.schema().fields()) {
+      if (struct.get(field) == null) {
+        data.add((Writable) null);
```
kkonstantine: nit: you don't need to cast

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
```diff
@@ -0,0 +1,122 @@
...
+  @Override
+  public String getExtension() {
+    return EXTENSION;
+  }
+
+  @Override
+  public RecordWriter getRecordWriter(final HdfsSinkConnectorConfig conf, final String filename) {
+    final Path path = new Path(filename);
+    return new RecordWriter() {
+      Writer writer = null;
```
kkonstantine: These are member variables and not local variables, thus assignment to `null` is redundant here.

src/main/java/io/confluent/connect/hdfs/orc/OrcHiveUtil.java
levzem: indeed it is :P time to purge the comments

src/main/java/io/confluent/connect/hdfs/orc/OrcHiveUtil.java
levzem: 1. yes
2. all HiveUtils behave like this

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
levzem: this follows how all the `RecordWriterProvider`s do it. Either we should change across all of them or none. Thoughts?

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
levzem: same comment as above

src/main/java/io/confluent/connect/hdfs/orc/OrcFileReader.java
levzem: This is called for every new file. Every other `FileReader` has it at info. Could be worth transitioning to `debug`

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
levzem: theres a `getExtension()` method and again follows the general principle lmao :P

src/test/java/io/confluent/connect/hdfs/orc/OrcDataFileReader.java
levzem: gross

src/main/java/io/confluent/connect/hdfs/orc/OrcFileReader.java
levzem: other `FileReader`s also missing

src/main/java/io/confluent/connect/hdfs/orc/OrcFileReader.java
levzem: ORC is a column file format so `Struct` is required

src/main/java/io/confluent/connect/hdfs/orc/OrcFileReader.java
levzem: this does not support nested `Struct` but will support a nested `Map` and `List`

src/main/java/io/confluent/connect/hdfs/orc/OrcHiveFactory.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  public HiveUtil createHiveUtil(
+      AbstractConfig conf,
+      io.confluent.connect.storage.hive.HiveMetaStore hiveMetaStore
+  ) {
+    return createHiveUtil((HdfsSinkConnectorConfig) conf, (HiveMetaStore) hiveMetaStore);
+  }
+
+  @Deprecated
```
kkonstantine: I haven't looked at the details too far, but I believe it'd be safe to skip the deprecated interfaces/methods. 

This format is introduced after the transition to storage-common, and therefore there's no obvious reason to retain compatibility with the deprecated stuff

levzem: fixed

src/main/java/io/confluent/connect/hdfs/orc/OrcUtil.java
```diff
@@ -0,0 +1,141 @@
...
+   */
+  public static Object[] convertStruct(Struct struct) {
+    List<Object> data = new LinkedList<>();
+    for (Field field : struct.schema().fields()) {
+      if (struct.get(field) == null) {
+        data.add(null);
+      } else {
+        Schema.Type schemaType = field.schema().type();
+        String name = field.name();
+        switch (schemaType) {
```
kkonstantine: This method with this long `switch` statement is a bit old-fashioned. 
It's also responsible for 3 checkstyle suppressions. 

Here's a suggestion that will work nicely instead can will make this code more readable and will need only one suppression (can't do much about `ClassDataAbstractionCoupling` - you need what you need). 

1. Declare an `EnumMap` as `private static EnumMap<Schema.Type, BiFunction<Struct, Field, Object>> conversions` 
2. Initialize it statically. E.g. 
```
static {
   conversions.put(BOOLEAN, OrcUtil::convertBoolean);
   ...
   conversions.put(MAP, OrcUtil::convertMap);
}
```
3. Discard your switch statement altogether and instead call: 
`data.add(conversions.get(schemaType).apply(struct, field));`
  (I assumed there's no type missing, but write appropriately if that's not the case)
4. Encapsulate the code for each conversion in the respective method. E.g. 
```
private static Object convertBoolean(Struct struct, Field field) {
    return new BooleanWritable(struct.getBoolean(field.name()));
  }
```
Remove the suppressions that are not required after the refactoring.

checkstyle/suppressions.xml
```diff
@@ -24,12 +24,12 @@
 
     <suppress
             checks="CyclomaticComplexity"
-            files="(DataWriter|HdfsSinkConnectorConfig|TopicPartitionWriter|WALFile).java"
+            files="(DataWriter|HdfsSinkConnectorConfig|TopicPartitionWriter|WALFile|OrcUtil).java"
```
kkonstantine: not required. See comment below

checkstyle/suppressions.xml
```diff
@@ -24,12 +24,12 @@
...
     />
 
     <suppress
             checks="JavaNCSS"
-            files="(DataWriter|TopicPartitionWriter).java"
+            files="(DataWriter|TopicPartitionWriter|OrcUtil).java"
```
kkonstantine: not required. See comment below

src/main/java/io/confluent/connect/hdfs/orc/OrcUtil.java
levzem: addressed

PR #501, opened 2020-06-18
CC-7360: Recover gracefully if closing a temporary file fails

levzem:
## Problem
 
Sometimes closing a temporary file fails, which means the file does not get flushed properly and therefore may result empty or corrupted files in HDFS showing up. 

## Solution

If a temporary file does not close the file, the connector will take the most conservative approach and will delete the temporary file attempt to rewrite all the records from scratch.

Note that this PR does not address closing temporary files when closing the `TopicPartitionWriter`. That should be addressed in a different PR.

- [x] yes
- [ ] no

##### If yes, where?

HDFS3 could face a similar issue

## Testing

Added a unit test that makes sure that the connector correctly handles the offsets and close failures. 

I also ran manual tests where I simulated closing failures (5% of the time) and see how the connector addresses them. They were a success.

Screen Shot 2020-06-19 at 2 22 13 PM

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release plan

Backporting to 5.0.x when this bug was introduced by #451 

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -751,8 +751,9 @@ public class TopicPartitionWriter {
         // still want to close all of the other data writers
         connectException = e;
         log.error(
-            "Failed to close temporary file for partition " + encodedPartition
-                + ". The connector will attempt to rewrite the temporary file."
+            "Failed to close temporary file for partition {}. The connector will attempt to"
+                + " rewrite the temporary file.",
+            encodedPartition
         );
       }
     }
@@ -765,7 +766,7 @@ public class TopicPartitionWriter {
         try {
           deleteTempFile(encodedPartition);
         } catch (ConnectException e) {
-          log.error("Failed to delete tmp file " + tempFiles.get(encodedPartition), e);
+          log.error("Failed to delete tmp file {}", tempFiles.get(encodedPartition), e);
         }
         startOffsets.remove(encodedPartition);
         offsets.remove(encodedPartition);
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -754,6 +754,9 @@ public class TopicPartitionWriter {
         buffer.clear();
       }
 
+      log.debug("Resetting offset for {} to {}", tp, offset);
+      context.offset(tp, offset);
+
       recordCounter = 0;
       throw connectException;
     }
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -738,6 +738,10 @@ public class TopicPartitionWriter {
       } catch (ConnectException e) {
         // still want to close all of the other data writers
         connectException = e;
+        log.error(
+            "Failed to close temporary file for partition " + encodedPartition
+                + ". The connector will attempt to rewrite the temporary file."
+        );
       }
     }
 
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -19,7 +19,6 @@ import org.apache.hadoop.fs.Path;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.errors.ConnectException;
-import org.apache.kafka.connect.errors.DataException;
 import org.apache.kafka.connect.errors.IllegalWorkerStateException;
 import org.apache.kafka.connect.errors.SchemaProjectorException;
 import org.apache.kafka.connect.sink.SinkRecord;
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -711,9 +711,35 @@ private void closeTempFile(String encodedPartition) {
...
+            "Failed to close temporary file for partition " + encodedPartition
+                + ". The connector will attempt to rewrite the temporary file."
+        );
+      }
+    }
+
+    if (connectException != null) {
+      // at least one tmp file did not close properly
+      for (String encodedPartition : tempFiles.keySet()) {
+        deleteTempFile(encodedPartition);
```
kkonstantine, line 732: what if delete fails with `ConnectException` too? Are we still going to have the same problem? Should we try to delete all the files at least?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -711,9 +711,35 @@ private void closeTempFile(String encodedPartition) {
   }
 
   private void closeTempFile() {
+    ConnectException connectException = null;
     for (String encodedPartition : tempFiles.keySet()) {
       // Close the file and propagate any errors
-      closeTempFile(encodedPartition);
+      try {
+        closeTempFile(encodedPartition);
```
kkonstantine, line 718: what about the other call to `closeTempFile(encodedPartition)`? Do we risk to leave stale tmp files in that case too?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -711,9 +711,35 @@ private void closeTempFile(String encodedPartition) {
...
+        connectException = e;
+        log.error(
+            "Failed to close temporary file for partition " + encodedPartition
+                + ". The connector will attempt to rewrite the temporary file."
+        );
+      }
+    }
+
+    if (connectException != null) {
+      // at least one tmp file did not close properly
```
kkonstantine, line 730: someone new to the code might wonder here, why don't we delete only the files that failed to close. 
IIUC that's because we won't be able to cherry-pick and skip the successful records on a retry (after we empty the buffer and reset the offsets). That would lead to duplicates. But should we add some info to this comment, so that we don't leave this completely to the reader's imagination?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
levzem: I am not sure what you mean. This guards against both calls and yes the empty file situation could occur in both situations (if that's what you are asking)

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
levzem: I have added more graceful delete however if it fails on a delete that shouldn't result in any empty files because the file will not be appended to the WAL so it will never be committed. Worst case that there will be some stale `tmp` files

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
kkonstantine: This method is called in `close` too

levzem: Yes and if it fails to close the file, it will delete it as per normal. There is no risk of leaving stale tmp files.

kkonstantine: From what I can see, `closeTempFile` and `deleteTempFile` are called within the same [try/catch](https://github.com/confluentinc/kafka-connect-hdfs/blob/b76d0422240594d0b5b619f7ff2b3e5b5f45d6ea/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L449-L450) block and if `closeTempFile` throws an exception, `deleteTempFile` will be skipped. If it's called later on again, I think it's worth mentioning what the actual workflow is, because it's not obvious.

levzem: yes I see what you mean now and yes that can happen and can leave stale temp files, but that logic is out of the scope of this PR. 

This PR is interested in the case when we try to close temp file so that we can commit it to HDFS and that operation fails. 

Closing the TPW gracefully to avoid stale temp files should be addressed in a different PR.

kkonstantine: Sorry, I missed the reply here. 

I don't think we are extending the scope too much if we add proper handling of the exception at that point too, given that we do that for this call here. Is there anything else that needs to happen besides deleting the file even in the presence of exception during close?

levzem: @kkonstantine I have added graceful handling of exceptions in `close()`

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -730,9 +742,41 @@ private void closeTempFile(String encodedPartition) {
...
     for (String encodedPartition : tempFiles.keySet()) {
       // Close the file and propagate any errors
-      closeTempFile(encodedPartition);
+      try {
+        closeTempFile(encodedPartition);
+      } catch (ConnectException e) {
+        // still want to close all of the other data writers
+        connectException = e;
+        log.error(
+            "Failed to close temporary file for partition " + encodedPartition
```
kkonstantine, line 754: nit: let's use `{}` for strings like we do elsewhere for consistency. In practice this message won't be skipped.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -730,9 +742,41 @@ private void closeTempFile(String encodedPartition) {
...
+
+    if (connectException != null) {
+      // at least one tmp file did not close properly therefore will try to recreate the tmp and
+      // delete all buffered records + tmp files and start over because otherwise there will be
+      // duplicates, since there is no way to reclaim the records in the tmp file.
+      for (String encodedPartition : tempFiles.keySet()) {
+        try {
+          deleteTempFile(encodedPartition);
+        } catch (ConnectException e) {
+          log.error("Failed to delete tmp file " + tempFiles.get(encodedPartition), e);
```
kkonstantine, line 768: same suggestion as above

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
levzem: oops nice catch!

levzem:
@kkonstantine 

yes depending on the partitioner the map may grow or not because it is a map of encoded partitions to temp files (default partitioner will only ever have 1 entry). A timestamp/field based partitioners are different stories. I would say this is not ideal behavior because we do not know when the TPW will be closed. IIUC the TPW can be open for the entire duration of the connectors runtime if no rebalancing occurs so we should definitely investigate this in another PR. It should be thread safe however, since TPW runs only on a per task basis.

PR #503, opened 2020-06-25
CC-9710: Retry effectively on exceptions from time-based commits

gharris1727:
## Problem
Exceptions thrown from time-based commits (in particular exceptions from appending to the WAL) aren't retried effectively, leading to corrupted files. 

* When an exception is thrown during time-based commits, the state machine is still updated to WRITE_STARTED
* When recovering in a subsequent non-empty write, the WRITE_STARTED state triggers the writer to start a new file, since the previous file was closed but not committed.
* This leads to the loss of the contents of the previous file, and corruption of the current file's offsets.

This test setup expects that the files 0-2, 3, 4-6 are created, but without the accompanying fix in the TopicPartitionWriter, the files 0-2 and 3-5 are created, and the 3-5 file is missing record 3, which was delivered in the initial batch and then overwritten.
## Solution

The time-based commits should keep the state-machine up-to-date, such that any recovery (non-empty or empty) will cleanly recover the file.

Also addresses minor bug where the memory format didn't properly overwrite data when the file was re-opened.

- [x] yes
- [ ] no

##### If yes, where?

This needs to be ported to HDFS3. https://confluentinc.atlassian.net/browse/CC-9759

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan
This will be backported to 5.0.x and released in the next patch releases of CP.
This change is safe to roll back, but data corruption as a result of the underlying bug will need manual remediation.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -420,14 +420,14 @@ public class TopicPartitionWriter {
           case WRITE_PARTITION_PAUSED:
             // committing files after waiting for rotateIntervalMs time but less than flush.size
             // records available
-            if (recordCounter > 0 && shouldRotateAndMaybeUpdateTimers(currentRecord, now)) {
-              log.info(
+            if (recordCounter == 0 || !shouldRotateAndMaybeUpdateTimers(currentRecord, now)) {
+              break;
+            } 
+            
+            log.info(
                   "committing files after waiting for rotateIntervalMs time but less than "
                       + "flush.size records available."
-              );
-            } else {
-              break;
-            }
+            );
             nextState();
           case SHOULD_ROTATE:
             updateRotationTimers(currentRecord);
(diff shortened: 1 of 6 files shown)


src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -412,24 +412,43 @@ public void write() {
...
+            // committing files after waiting for rotateIntervalMs time but less than flush.size
+            // records available
+            if (recordCounter > 0 && shouldRotateAndMaybeUpdateTimers(currentRecord, now)) {
+              log.info(
+                  "committing files after waiting for rotateIntervalMs time but less than "
+                      + "flush.size records available."
+              );
+            } else {
+              break;
+            }
```
levzem, line 430:
```suggestion
            // committing files after waiting for rotateIntervalMs time but less than flush.size
            // records available
            if (recordCounter == 0 || !shouldRotateAndMaybeUpdateTimers(currentRecord, now)) {
              break;
            } 
            
            log.info(
                  "committing files after waiting for rotateIntervalMs time but less than "
                      + "flush.size records available."
            );
```

nit: a little a bit cleaner

PR #509, opened 2020-07-09
CC-8704: Resolve jetty CVE

dosvath:
## Problem
This PR addresses a CVE caused by `jetty-all 7.6.0.v20120127`. 

## Solution
Exclude the `jetty-all` dependency that is brought in by `kafka-connect-storage-hive`, same solution as in [gcp-dataproc](https://github.com/confluentinc/kafka-connect-gcp-dataproc/blob/master/pom.xml#L353-L356) and [object-store-source](https://github.com/confluentinc/kafka-connect-object-store-source/blob/master/sink-integration-tests/pom.xml#L190-L193) connectors. 

- [X] yes
- [ ] no

##### If yes, where?
All connectors using `jetty-all 7.6.0.v20120127`.

##### Testing done:
- [X] Unit tests
- [x] Integration tests
- [x] System tests
- [ ] Manual tests

## Release Plan

 
Backporting to 5.4.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,6 +108,12 @@
             <groupId>io.confluent</groupId>
             <artifactId>kafka-connect-storage-hive</artifactId>
             <version>${confluent.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.eclipse.jetty.aggregate</groupId>
+                    <artifactId>jetty-all</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>com.github.spotbugs</groupId>


dosvath:
> if we do not need `jetty-all` we should remove it from `storage-hive` and not every individual connector

@levzem I had that thought as well but considered the possibility that some connectors may need it. It looks like it's being removed in many connector as well, I would support the approach to remove it from `storage-hive` in general. cc @kkonstantine

dosvath:
Update: muckrake [system tests pass](https://jenkins.confluent.io/job/system-test-confluent-platform-branch-builder/4053/).

dosvath:
@levzem @kkonstantine A thought before removing from `storage-hive`: 
Would adding the exclusions to `storage-hive` require us to upgrade the version of `storage-hive` in all affected connectors? In that case, we should consider if the updated version includes other changes that may cause compatibility issues.

levzem:
> @levzem @kkonstantine A thought before removing from `storage-hive`:
> Would adding the exclusions to `storage-hive` require us to upgrade the version of `storage-hive` in all affected connectors? In that case, we should consider if the updated version includes other changes that may cause compatibility or build issues.

if we backport the exclusion, then we should not be having any problems

kkonstantine:
> @levzem @kkonstantine A thought before removing from `storage-hive`:
> Would adding the exclusions to `storage-hive` require us to upgrade the version of `storage-hive` in all affected connectors? In that case, we should consider if the updated version includes other changes that may cause compatibility or build issues.

Good point @dosvath. Let's inherit the inclusion from `storage-common` to the connectors that move in lock step with this repo, and let's add an exclusion to the rest if we don't want to wait. Upgrading to a bugfix of `storage-common` should not be an issue because it's only a patch release.

dosvath:
@kkonstantine Sounds good, pending approval.

kkonstantine:
@dosvath this connector moves in lockstep with `storage-common`. So it's fine to get the fix from there and close the PR here.

dosvath:
@kkonstantine [made PR](https://github.com/confluentinc/kafka-connect-storage-common/pull/139) to `storage-common`, closing this PR here.

PR #336, opened 2018-06-13
Fix parquet hive duplicate column

RussWinkler:
Fix an issue where the Parquet Hive Utility would add the partition column as a regular column, then allow hive to create an additional partition column thereby creating a duplicate column for the partition column with the same name and creating an invalid hive table definition.  

Updated the associated tests accordingly.

This is an updated fix for my older submitted issue: https://github.com/confluentinc/kafka-connect-hdfs/issues/238

Issue #238, opened 2017-10-02
Ability to rename partition column for partitioners


RussWinkler:
When you enable hive integration with a connector and use the FieldPartitioner, it creates a new column to partition the data on with the same name as the column that are you partitioning on.  This was brought up previously in https://github.com/confluentinc/kafka-connect-hdfs/issues/221 but seems to have been satisfactorily answered by the original author.  

Our problem is that we are using Presto to connect to these hive tables and Presto chokes when there are two columns with the same name in the hive table:

`Query 20170314_185633_00138_fnxpw failed: Hive metadata for table test_table is invalid: Table descriptor contains duplicate columns`

It seems like it would be pretty easy to allow a connector parameter to let the hive metadata know what you want the partition column to be called (maybe by overloading the path.format parameter?).  At the very least add something to the column name so the partition column name isn't _exactly_ the same as the original column?

Thanks!

GeryD:
I'm have this issue too.
From  [hive documentation](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-Create/Drop/Alter/UseDatabase) (§Partitioned Tables), field name use for partitioning should be renamed in table or should not be present

justpresident:
The fix above is for parquet onlu, but the same problem appears for other formats as well. Specifically I have it with avro and while the table is readable by beeline client, spark 2.2.0 fails with duplicate column error.

RussWinkler:
@GeryD and @justpresident I have fixed this issue in the most most recent versions of kafka-connect-hdfs and have opened a pull request with confluent to get it merged.  I would love to get this PR accepted before the next release so if you would please ask for it to be reviewed over there I would appreciate it!

https://github.com/confluentinc/kafka-connect-hdfs/pull/336

@justpresident I didn't know Avro was having a similar issue but you can probably follow the same format to also fix Avro!

GeryD:
Hi @RussWinkler,
Unfortunately I'm using avro too

Titiabackup:
Hi,

Do you know if this issue have a fix in a version ?
Thanks

abhisheksahani:
Hi @RussWinkler  I followed your changes and made the exact same changes to ParquetHiveUtil 5.2.2 snapshot
but for me its not working can you please look at the code.

 @Override
  public void createTable(
      String database,
      String tableName,
      Schema schema,
      Partitioner partitioner
  ) throws HiveMetaStoreException {
[...]

Issue #221, opened 2017-08-17
duplicate column in hive table when use partition.field.name defined in  avro schema


duke2012:
Below items is set in quickstart-hdfs.properties file  where partition_date is a field defined in avro schema,everything is ok except this field duplicate in hive table. anyone know why? thanks in advance.

`partitioner.class=io.confluent.connect.hdfs.partitioner.FieldPartitioner
partition.field.name=partition_date`

**And hive table structure as below:**

latest_search_keyword	string              	                    
referrer            	string              	                    
referrer_host       	string              	                    
is_first_day        	boolean             	                    
is_first_time       	boolean             	                    
partition_date      	string              	                    
partition_date      	string              	                    
	 	 
Partition Information	 	 
col_name            	data_type           	comment             
partition_date      	string

lomignet:
`partition.field.name` will add a pseudo-column named partition_date (in your example) holding the partition value. This is how Hive handles partitions. As this column already exists in your data, you end up having a duplicated column.

The solutions could be:
- choose another name for `partition.field.name`,
- choose another name in your avro schema for partition_date,
- remove partition_date from your schema if your goal was to have it filled by he connector, as it is not how it works.

duke2012:
@lomignet  thanks ,it saves me a day.

RussWinkler:
I know this thread is a bit old, but I'm confused by the answer here.  I was under the impression that `partition.field.name` indicated which field the Partitioner should use, not the name of the field that the Partitioner will create for partitioning.  If you choose another name for `partition.field.name` how will the Partitioner know which field in the avro schema to partition the data on?

lomignet:
@RussWinkler You are very right. Choosing another name is relevant for a `TimebasedPartitioner` where you can specify your own field names for the different time elements via `path.format` but does not make sense for a `FieldPartitioner`.

levzem:
closing as this solution was provided

kkonstantine:
Fixed by: https://github.com/confluentinc/kafka-connect-hdfs/pull/526 and will appear in 10.0.0 and newer versions.

PR #513, opened 2020-08-18
Fix for hive parquet duplicate columns while using field partitioner

kaushiksrinivas:
## Problem
field partitioner creates duplicate hive parquet partition columns.

## Solution
Check if the column is a partition column and ignore them while creating the table.

references:
https://github.com/confluentinc/kafka-connect-hdfs/pull/336/files
https://github.com/confluentinc/kafka-connect-hdfs/issues/238
https://github.com/confluentinc/kafka-connect-hdfs/issues/221

- [ ] yes
- [ ] no

## Test Strategy
unit tests updated accordingly.

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java
+++ b/src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java
@@ -54,6 +54,15 @@ public class ParquetHiveUtil extends HiveUtil {
   public void alterSchema(String database, String tableName, Schema schema) {
     Table table = hiveMetaStore.getTable(database, tableName);
     List<FieldSchema> columns = HiveSchemaConverter.convertSchema(schema);
+    List<FieldSchema> partitions = table.getPartCols();
+    for (FieldSchema partition : partitions) {
+      for (FieldSchema column : columns) {
+        if (partition.getName().equals(column.getName())) {
+          columns.remove(column);
+          break;
+        }
+      }
+    }
     table.setFields(columns);
     hiveMetaStore.alterTable(table);
   }
@@ -79,8 +88,17 @@ public class ParquetHiveUtil extends HiveUtil {
     }
     // convert Connect schema schema to Hive columns
     List<FieldSchema> columns = HiveSchemaConverter.convertSchema(schema);
+    List<FieldSchema> partitions = partitioner.partitionFields();
+    for (FieldSchema partition : partitions) {
+      for (FieldSchema column : columns) {
+        if (partition.getName().equals(column.getName())) {
+          columns.remove(column);
+          break;
+        }
+      }
+    }
     table.setFields(columns);
-    table.setPartCols(partitioner.partitionFields());
+    table.setPartCols(partitions);
     return table;
   }
 
--- a/src/test/java/io/confluent/connect/hdfs/parquet/HiveIntegrationParquetTest.java
+++ b/src/test/java/io/confluent/connect/hdfs/parquet/HiveIntegrationParquetTest.java
@@ -156,6 +156,7 @@ public class HiveIntegrationParquetTest extends HiveTestBase {
     DataWriter hdfsWriter = new DataWriter(connectorConfig, context, avroData);
 
     Schema schema = createSchema();
+	Field partitionField = schema.field("int");
     List<Struct> records = createRecordBatches(schema, 3, 3);
     List<SinkRecord> sinkRecords = createSinkRecords(records, schema);
 
@@ -167,7 +168,8 @@ public class HiveIntegrationParquetTest extends HiveTestBase {
 
     List<String> expectedColumnNames = new ArrayList<>();
     for (Field field : schema.fields()) {
-      expectedColumnNames.add(field.name());
+      if(!field.name().equals(partitionField.name()))
+        expectedColumnNames.add(field.name());
     }
 
     List<String> actualColumnNames = new ArrayList<>();
@@ -199,7 +201,8 @@ public class HiveIntegrationParquetTest extends HiveTestBase {
       for (int j = 0; j < 3; ++j) {
         List<String> result = new ArrayList<>();
         for (Field field : schema.fields()) {
-          result.add(String.valueOf(records.get(i).get(field.name())));
+          if(!field.name().equals(partitionField.name()))
+            result.add(String.valueOf(records.get(i).get(field.name())));
         }
         expectedResults.add(result);
       }
@@ -233,6 +236,10 @@ public class HiveIntegrationParquetTest extends HiveTestBase {
         .field("country", Schema.STRING_SCHEMA)
         .field("state", Schema.OPTIONAL_STRING_SCHEMA)
         .build();
+		
+	List<String> partitionFieldNames = new ArrayList<>();
+    partitionFieldNames.add("country");
+    partitionFieldNames.add("state");
 
     List<Struct> records = Arrays.asList(
         new Struct(schema)
@@ -258,7 +265,8 @@ public class HiveIntegrationParquetTest extends HiveTestBase {
 
     List<String> expectedColumnNames = new ArrayList<>();
     for (Field field : schema.fields()) {
-      expectedColumnNames.add(field.name());
+      if(!partitionFieldNames.contains(field.name()))
+        expectedColumnNames.add(field.name());
     }
 
     List<String> actualColumnNames = new ArrayList<>();
@@ -278,7 +286,7 @@ public class HiveIntegrationParquetTest extends HiveTestBase {
     assertEquals(expectedPartitions, partitions);
 
     List<List<String>> expectedResults = Arrays.asList(
-        Arrays.asList("1", "mx", "NULL"),
+        Arrays.asList("1", "mx", "null"),
         Arrays.asList("1", "us", "ca"),
         Arrays.asList("1", "us", "tx")
     );


CsatariGergely:
@kkonstantine, @dosvath, @levzem can you please help to find the correct reviewers for this pr? Thanks.

cdpop:
This is actually a feature where we have 2 customers asking for this:
https://confluentinc.atlassian.net/browse/FF-2986

Do we have any updates on this?

CsatariGergely:
@cdpop we are waiting for a review of someone with approving rights. @kkonstantine can you please approve or provide comments?

src/test/java/io/confluent/connect/hdfs/parquet/HiveIntegrationParquetTest.java
```diff
@@ -258,7 +265,8 @@ public void testHiveIntegrationFieldPartitionerParquetMultiple() throws Exceptio
 
     List<String> expectedColumnNames = new ArrayList<>();
     for (Field field : schema.fields()) {
-      expectedColumnNames.add(field.name());
+      if(!partitionFieldNames.contains(field.name()))
+        expectedColumnNames.add(field.name());
```
dosvath, line 269:
```suggestion
      if(!partitionFieldNames.contains(field.name())) {
        expectedColumnNames.add(field.name());
      }
```

src/test/java/io/confluent/connect/hdfs/parquet/HiveIntegrationParquetTest.java
```diff
@@ -233,6 +236,10 @@ public void testHiveIntegrationFieldPartitionerParquetMultiple() throws Exceptio
         .field("country", Schema.STRING_SCHEMA)
         .field("state", Schema.OPTIONAL_STRING_SCHEMA)
         .build();
+		
+	List<String> partitionFieldNames = new ArrayList<>();
```
dosvath, line 240: fix indentation

src/test/java/io/confluent/connect/hdfs/parquet/HiveIntegrationParquetTest.java
```diff
@@ -199,7 +201,8 @@ public void testHiveIntegrationFieldPartitionerParquet() throws Exception {
       for (int j = 0; j < 3; ++j) {
         List<String> result = new ArrayList<>();
         for (Field field : schema.fields()) {
-          result.add(String.valueOf(records.get(i).get(field.name())));
+          if(!field.name().equals(partitionField.name()))
+            result.add(String.valueOf(records.get(i).get(field.name())));
```
dosvath, line 205:
```suggestion
          if(!field.name().equals(partitionField.name())) {
            result.add(String.valueOf(records.get(i).get(field.name())));
          }
```

src/test/java/io/confluent/connect/hdfs/parquet/HiveIntegrationParquetTest.java
```diff
@@ -156,6 +156,7 @@ public void testHiveIntegrationFieldPartitionerParquet() throws Exception {
     DataWriter hdfsWriter = new DataWriter(connectorConfig, context, avroData);
 
     Schema schema = createSchema();
+	Field partitionField = schema.field("int");
```
dosvath, line 159: fix indentation

src/test/java/io/confluent/connect/hdfs/parquet/HiveIntegrationParquetTest.java
```diff
@@ -167,7 +168,8 @@ public void testHiveIntegrationFieldPartitionerParquet() throws Exception {
 
     List<String> expectedColumnNames = new ArrayList<>();
     for (Field field : schema.fields()) {
-      expectedColumnNames.add(field.name());
+      if(!field.name().equals(partitionField.name()))
+        expectedColumnNames.add(field.name());
```
dosvath, line 172:
```suggestion
      if(!field.name().equals(partitionField.name())) {
        expectedColumnNames.add(field.name());
      }
```

src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java
```diff
@@ -79,8 +88,17 @@ private Table constructParquetTable(
...
     List<FieldSchema> columns = HiveSchemaConverter.convertSchema(schema);
+    List<FieldSchema> partitions = partitioner.partitionFields();
+    for (FieldSchema partition : partitions) {
+      for (FieldSchema column : columns) {
+        if (partition.getName().equals(column.getName())) {
+          columns.remove(column);
+          break;
+        }
+      }
+    }
```
dosvath, line 99: Adding a helper method for this would be great so that we can avoid copying the code on the two instances.

dosvath (commented):
Looks good overall, some cleanup suggestions. @CsatariGergely @kaushiksrinivas has there been any manual/integration verification done on hive for the fix?

dosvath:
Hi @CsatariGergely, @kaushiksrinivas I wasn't able to fully replicate the issue using Hive and `parquet` format. Could you provide some more details about the Hive/HDFS versions, connector version,  and an example schema/record that causes the issue along with the `partition.field.name` that you set? These would be really useful in replicating the issue and verifying that the fix in this PR works. For my testing I used connector version `5.5.2`, hive `2.3.7`, hadoop `2.10.1`, `"partition.field.name": "firstname"` and I was using the types of records below using the avro-console producer:
```
"key1":{"firstname": "Daniel", "lastname":"O", "website":"confluent.io"}
```
When listing the table in hive I only see one instance of each column. I inspected the parquet file and I couldn't see duplicate columns either. Does the issue occur when the connector creates the table, or only when the table exists before starting the connector, or both?

dosvath:
I have looked into the unit tests on this PR and they currently do not verify the correction of duplicate column functionality. With the above changes the tests remove a unique column from the schema and assert on the expected columns without the field partitioned column. Specifically the [createSchema()](https://github.com/confluentinc/kafka-connect-storage-common/blob/2edc0df7f48ed0b0c22a5434c765c24cff3d5369/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java#L63) method gives us a schema with `bool`,`int`, `long`, `double` and `float`. After the PR changes the expected columns become `bool`, `long`, `double` and `float`. This is an issue because that way the tests aren't confirming the removal of a duplicate column but of a unique column, and assumes that hive will always create a column for the partitioned field.  

This also leads me to believe more that we are seeing an issue due to a certain hive version that creates an additional column for the field partitioned column, but the connector is not creating this duplicate column, so I don't believe we have the correct approach on the unit tests. 

Regarding the functional changes, as we saw for my above runs the hive version did not create a duplicate column, so dropping the field partitioned column by default may result in the entire omission of the column in hive versions which do not create a partitioned field column.

kaushiksrinivas:
> Looks good overall, some cleanup suggestions. @CsatariGergely @kaushiksrinivas has there been any manual/integration verification done on hive for the fix?

Yes Manual verification is done for the same.

kaushiksrinivas:
> I have looked into the unit tests on this PR and they currently do not verify the correction of duplicate column functionality. With the above changes the tests remove a unique column from the schema and assert on the expected columns without the field partitioned column. Specifically the [createSchema()](https://github.com/confluentinc/kafka-connect-storage-common/blob/2edc0df7f48ed0b0c22a5434c765c24cff3d5369/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java#L63) method gives us a schema with `bool`,`int`, `long`, `double` and `float`. After the PR changes the expected columns become `bool`, `long`, `double` and `float`. This is an issue because that way the tests aren't confirming the removal of a duplicate column but of a unique column, and assumes that hive will always create a column for the partitioned field.

Hi @dosvath 
This is done like this, since the createSchema() is used commonly across all tests and this is applied to parquet only.
Can you suggest any other options ? or we retain this for parquet tests only ?
Thanks,
Kaushik.

dosvath:
Hi @kaushiksrinivas @CsatariGergely, thank you for your efforts on this, we found the approach in this PR really helpful. We'll be moving forward with this fix in this PR https://github.com/confluentinc/kafka-connect-hdfs/pull/526 that also includes fixes for Avro.

dosvath:
> > I have looked into the unit tests on this PR and they currently do not verify the correction of duplicate column functionality. With the above changes the tests remove a unique column from the schema and assert on the expected columns without the field partitioned column. Specifically the [createSchema()](https://github.com/confluentinc/kafka-connect-storage-common/blob/2edc0df7f48ed0b0c22a5434c765c24cff3d5369/core/src/main/java/io/confluent/connect/storage/StorageSinkTestBase.java#L63) method gives us a schema with `bool`,`int`, `long`, `double` and `float`. After the PR changes the expected columns become `bool`, `long`, `double` and `float`. This is an issue because that way the tests aren't confirming the removal of a duplicate column but of a unique column, and assumes that hive will always create a column for the partitioned field.
> 
> Hi @dosvath
> This is done like this, since the createSchema() is used commonly across all tests and this is applied to parquet only.
> Can you suggest any other options ? or we retain this for parquet tests only ?
> Thanks,
> Kaushik.

The solution ended up being to use `table.getAllCols()` in the tests which also revealed the duplicate columns. [See here](https://github.com/confluentinc/kafka-connect-hdfs/pull/526/files#diff-89d2630c60c9e8951ced0329a4eb989322a0389a551e56413da3d0dc7d07edb1R239).

kkonstantine:
Closing in favor of: https://github.com/confluentinc/kafka-connect-hdfs/pull/526 
which applied the fix for both parquet and avro. Thanks for the contribution @kaushiksrinivas !

kaushiksrinivas:
> Hi @kaushiksrinivas @CsatariGergely, thank you for your efforts on this, we found the approach in this PR really helpful. We'll be moving forward with this fix in this PR #526 that also includes fixes for Avro.

Sure. thanks

kaushiksrinivas:
> Closing in favor of: #526
> which applied the fix for both parquet and avro. Thanks for the contribution @kaushiksrinivas !

Sure. thanks.

Issue #515, opened 2020-09-09
Exactly-once should also work without topic name included in the path 


hariprasad-k:
This change is necessary to retain EOS semantics to support flexible storage partitioning scheme as proposed here: https://github.com/confluentinc/kafka-connect-storage-common/issues/126

JozoVilcek:
@hariprasad-k this looks like something I am looking for too. What is holding the progress here? Did you manage to get this done somehow else?

hariprasad-k:
> @hariprasad-k this looks like something I am looking for too. What is holding the progress here? Did you manage to get this done somehow else?

@JozoVilcek  Yes, if you apply the below patches, it does work for us to exclude topic names with EoS semantics. We are still awaiting for committers to review this change proposal though, so that we don’t need to maintain this internal fork.

https://github.com/confluentinc/kafka-connect-storage-common/pull/127/files
https://github.com/confluentinc/kafka-connect-hdfs/pull/528/files

PR #516, opened 2020-09-09
Recover the offset from HDFS, even if topic name is not present in storage path

hariprasad-k:
#346  Problem
Exactly-once semantics should also work without topic name included in the path.

## Solution
The function that recover offsets from file names will use the correct path by taking into account a new configuration flag `path_include_topicname ` introduced with confluentinc/kafka-connect-storage-common#126.

Further, the backward compatibility of appending the topic name is retained by default value `true` for `path_include_topicname`.

- [x] yes
- [ ] no

##### If yes, where?
This change is necessary to retain EOS semantics to support flexible storage partitioning scheme as proposed here: confluentinc/kafka-connect-storage-common#126

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/FileUtils.java
+++ b/src/main/java/io/confluent/connect/hdfs/FileUtils.java
@@ -31,6 +31,7 @@ import java.util.regex.Matcher;
 
 import io.confluent.connect.hdfs.filter.CommittedFileFilter;
 import io.confluent.connect.hdfs.storage.Storage;
+import io.confluent.connect.storage.common.StorageCommonConfig;
 
 public class FileUtils {
   private static final Logger log = LoggerFactory.getLogger(FileUtils.class);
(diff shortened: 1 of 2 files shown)


hariprasad-k:
Fixes #515

src/main/java/io/confluent/connect/hdfs/FileUtils.java
```diff
@@ -101,7 +102,12 @@ public static String committedFileName(
   }
 
   public static String topicDirectory(String url, String topicsDir, String topic) {
-    return url + "/" + topicsDir + "/" + topic;
+    boolean topicInPath = Boolean.valueOf(StorageCommonConfig.PATH_INCLUDE_TOPICNAME_CONFIG);
```
OneCricketeer, line 105: This will always be false since you're parsing the config string, not looking up the value in the config

OneCricketeer:
Could you please add unit tests?

PR #327, opened 2018-05-12
Fix #326 by using config's avroDataConfig method

dcsobral:
While there's nothing wrong in this change itself, `avroDataConfig` has an issue, reported on confluentinc/kafka-connect-storage-common#64 (pull request with fix has been provided), that makes the resulting `AvroDataConfig` not have the cache size, which is the _only_ configuration provided before this PR.

As long as that is fixed, this preserves the cache information _and_ provides the other settings.

PR #522, opened 2020-10-20
Fix #326 by using config's avroDataConfig method

### LINKED ISSUE #326: AvroData is initialized without configuration
On `AvroFormat` and a couple of other places, `AvroData` is instantiated with a sole hard-coded cache size parameter, ignoring the other `AvroDataConfig` parameters. In particular, the absence of `ENHANCED_AVRO_SCHEMA_SUPPORT_CONFIG` can prevent it from encoding the data generated by `AvroConverter`.

To compound this, `StorageSinkConnectorConfig`, and, therefore, `HdfsSinkConnectorConfig`, actually has a method that returns `AvroDataConfig`, so code to just pass all its parameters is shorter than what is being done right now.

Prathibha-m:
## Problem
HDFS connector configurations, specifically connect.meta.data and enhanced.avro.schema.support essentially NOOP when configured; they are ignored when AvroData is configured.

## Solution
Use fix from https://github.com/confluentinc/kafka-connect-hdfs/pull/327 

- [x] yes Other blob store connectors (merged)
- [ ] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -75,10 +75,7 @@ public class HdfsSinkTask extends SinkTask {
         DateTimeZone.forID(timeZoneString);
       }
 
-      int schemaCacheSize = connectorConfig.getInt(
-          HdfsSinkConnectorConfig.SCHEMA_CACHE_SIZE_CONFIG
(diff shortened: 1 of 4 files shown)


levzem (approved):
LGTM

PR #526, opened 2020-10-29
CC-12567: Fix duplicate columns in Hive for Parquet and Avro formats

dosvath:
## Problem
Duplicate columns are created when the field partitioner is used. 

## Solution
Remove duplicate column instances.

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [X] Manual tests

## Release Plan

 
Could not backport to a earlier branch as MiniDFSCluster tests are unstable on older branches.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
+++ b/src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
@@ -70,7 +70,7 @@ public class AvroHiveUtil extends HiveUtil {
       Schema schema
   ) throws HiveMetaStoreException {
     Table table = hiveMetaStore.getTable(database, tableName);
-    Schema filteredSchema = filterSchema(schema, table.getPartitionKeys());
+    Schema filteredSchema = excludePartitionFieldsFromSchema(schema, table.getPartitionKeys());
     table.getParameters().put(AVRO_SCHEMA_LITERAL,
         avroData.fromConnectSchema(filteredSchema).toString());
     hiveMetaStore.alterTable(table);
@@ -96,7 +96,7 @@ public class AvroHiveUtil extends HiveUtil {
     } catch (HiveException e) {
       throw new HiveMetaStoreException("Cannot find input/output format:", e);
     }
-    Schema filteredSchema = filterSchema(schema, partitioner.partitionFields());
+    Schema filteredSchema = excludePartitionFieldsFromSchema(schema, partitioner.partitionFields());
     List<FieldSchema> columns = HiveSchemaConverter.convertSchema(filteredSchema);
     table.setFields(columns);
     table.setPartCols(partitioner.partitionFields());
@@ -109,18 +109,18 @@ public class AvroHiveUtil extends HiveUtil {
    * Remove the column(s) that is later re-created by Hive when using the
    * {@code partition.field.name} config.
    *
-   * @param oldSchema the old schema to remove fields from
+   * @param originalSchema the old schema to remove fields from
    * @param partitionFields the fields used for partitioning
    * @return the new schema without the fields used for partitioning
    */
-  private Schema filterSchema(Schema oldSchema, List<FieldSchema> partitionFields) {
+  private Schema excludePartitionFieldsFromSchema(Schema originalSchema, List<FieldSchema> partitionFields) {
     Set<String> partitions = partitionFields.stream()
         .map(FieldSchema::getName).collect(Collectors.toSet());
 
     SchemaBuilder newSchema = SchemaBuilder.struct();
-    for (Field f : oldSchema.fields()) {
-      if (!partitions.contains(f.name())) {
-        newSchema.field(f.name(), f.schema());
+    for (Field field : originalSchema.fields()) {
+      if (!partitions.contains(field.name())) {
+        newSchema.field(field.name(), field.schema());
       }
     }
     return newSchema;
--- a/src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
+++ b/src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
@@ -70,7 +70,7 @@ public class AvroHiveUtil extends HiveUtil {
       Schema schema
   ) throws HiveMetaStoreException {
     Table table = hiveMetaStore.getTable(database, tableName);
-    Schema filteredSchema = filteredSchema(schema, table.getPartitionKeys());
+    Schema filteredSchema = filterSchema(schema, table.getPartitionKeys());
     table.getParameters().put(AVRO_SCHEMA_LITERAL,
         avroData.fromConnectSchema(filteredSchema).toString());
     hiveMetaStore.alterTable(table);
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
```diff
@@ -91,9 +97,49 @@ private Table constructAvroTable(
...
+  /**
+   * Remove the column(s) that is later re-created by Hive when using the
+   * {@code partition.field.name} config.
+   *
+   * @param columns the hive columns from
+   *                {@link HiveSchemaConverter#convertSchema(Schema) convertSchema}.
+   * @param partitionFields the fields used for partitioning
+   */
+  private void removeFieldPartitionColumn(List<FieldSchema> columns,
+      List<FieldSchema> partitionFields) {
```
levzem, line 139:
```suggestion
  private void removeFieldPartitionColumn(
      List<FieldSchema> columns,
      List<FieldSchema> partitionFields
   ) {
```

src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java
```diff
@@ -119,4 +123,20 @@ private String getHiveParquetSerde() {
...
+  /**
+   * Remove the column that is later re-created by Hive when using the
+   * {@code partition.field.name} config.
+   *
+   * @param columns the hive columns from
+   *                {@link HiveSchemaConverter#convertSchema(Schema) convertSchema}.
+   * @param partitionFields the fields used for partitioning
+   */
+  private void removeFieldPartitionColumn(List<FieldSchema> columns,
+      List<FieldSchema> partitionFields) {
```
levzem, line 136:
```suggestion
  private void removeFieldPartitionColumn(
      List<FieldSchema> columns,
      List<FieldSchema> partitionFields
  ) {
```

src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
```diff
@@ -91,9 +97,49 @@ private Table constructAvroTable(
...
+    Set<String> partitions = partitionFields.stream()
+        .map(FieldSchema::getName).collect(Collectors.toSet());
+
+    columns.removeIf(column -> partitions.contains(column.getName()));
+  }
```
levzem, line 144: we can move this method to `HiveUtil` instead of duplicating it across two classes

src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java
```diff
@@ -79,6 +82,7 @@ private Table constructParquetTable(
     }
     // convert Connect schema schema to Hive columns
     List<FieldSchema> columns = HiveSchemaConverter.convertSchema(schema);
```
levzem, line 84: we dont need to filter the schema here?

src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
dosvath: I held off due to `HiveUtil` being `@Deprecated` and with the note 
```
// NOTE: DO NOT add or modify this class as it is maintained for compatibility
```

dosvath: Would be great indeed if we could add it there, wdyt @kkonstantine ?

src/main/java/io/confluent/connect/hdfs/parquet/ParquetHiveUtil.java
dosvath: Not in this case, as the `schema` is not used later like in the `AvroHiveUtil` for `table.getParameters()`. Now that you mention it though, it looks like we could simply filter the schema and then we don't need the `removeFieldPartitionColumn` to call on the columns.

src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
levzem: would have to add it here https://github.com/confluentinc/kafka-connect-storage-common/blob/master/hive/src/main/java/io/confluent/connect/storage/hive/HiveUtil.java then

levzem: considering there the future work planned, im fine leaving as is

levzem (approved):
LGTM

src/test/java/io/confluent/connect/hdfs/avro/HiveIntegrationAvroTest.java
```diff
@@ -239,17 +230,20 @@ public void testHiveIntegrationFieldPartitionerAvro() throws Exception {
...
     List<String> expectedColumnNames = new ArrayList<>();
-    for (Field field: schema.fields()) {
+    for (Field field : schema.fields()) {
       expectedColumnNames.add(field.name());
     }
+    Collections.sort(expectedColumnNames);
 
     List<String> actualColumnNames = new ArrayList<>();
-    for (FieldSchema column: table.getSd().getCols()) {
+    for (FieldSchema column : table.getAllCols()) {
```
kkonstantine, line 239: let's add a comment here, since you discovered that this is required to surface duplicate columns.

src/test/java/io/confluent/connect/hdfs/avro/HiveIntegrationAvroTest.java
```diff
@@ -352,32 +347,28 @@ public void testHiveIntegrationFieldPartitionerAvroMultiple() throws Exception {
...
+            hiveMetaStore.tableNameConverter(TOPIC)
     );
     String[] rows = result.split("\n");
-    assertEquals(9, rows.length);
+    assertEquals(expectedResults.size(), rows.length);
     for (int i = 0; i < rows.length; ++i) {
       String[] parts = HiveTestUtils.parseOutput(rows[i]);
-      for (int j = 0; j < expectedResult.get(i).length; ++j) {
-        assertEquals(expectedResult.get(i)[j], parts[j]);
+      int j = 0;
```
kkonstantine, line 369: If you need the j variable the classic for loop was fine.

src/test/java/io/confluent/connect/hdfs/parquet/HiveIntegrationParquetTest.java
```diff
@@ -169,11 +172,14 @@ public void testHiveIntegrationFieldPartitionerParquet() throws Exception {
     for (Field field : schema.fields()) {
       expectedColumnNames.add(field.name());
     }
+    Collections.sort(expectedColumnNames);
 
     List<String> actualColumnNames = new ArrayList<>();
-    for (FieldSchema column : table.getSd().getCols()) {
+    for (FieldSchema column : table.getAllCols()) {
```
kkonstantine, line 178: same as above

src/test/java/io/confluent/connect/hdfs/parquet/HiveIntegrationParquetTest.java
```diff
@@ -260,11 +270,14 @@ public void testHiveIntegrationFieldPartitionerParquetMultiple() throws Exceptio
...
```
kkonstantine, line 276: same as above

kkonstantine (commented):
Thanks @dosvath !
Looks pretty good overall. Left a just few comments.

kkonstantine (approved):
LGTM
Nice work @dosvath !

One last thing, is that we should target 10.0.x

PR #528, opened 2020-11-03
Recover the offset from HDFS, even if topic name is not present in storage path

hariprasad-k:
## Problem
Exactly-once semantics should also work without topic name included in the path.

## Solution
The function that recover offsets from file names will use the correct path by taking into account a new configuration flag `path_include_topicname ` introduced with confluentinc/kafka-connect-storage-common#126.

Further, the backward compatibility of appending the topic name is retained by default value `true` for `path_include_topicname`.

- [x] yes
- [ ] no

##### If yes, where?
This change is necessary to retain EOS semantics to support flexible storage partitioning scheme as proposed here: confluentinc/kafka-connect-storage-common#126

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -399,7 +399,12 @@ public class DataWriter {
 
     try {
       for (String topic : topics) {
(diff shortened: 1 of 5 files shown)


hariprasad-k:
Fixes #515

PR #531, opened 2020-12-01
HCD-481 : Adding reconnection logic added for HDFS2 sink connector

SanchayGupta1197:
## Problem
The connector was failing for `Avro` and `ORC` format when the connection to the cluster was interrupted.

## Solution
**Fix for Avro**
- The schema was re-initialized to `null` in order to invalidate the current temp file.
- The `AvroRuntimeException` was caught and thrown as a `ConnectException`.

**Fix for ORC**
- The schema was re-initialized to `null` in order to invalidate the current temp file.
- The `AlreadyBeingCreatedException` was handled by deleting the temp file from the cluster.

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
+++ b/src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
@@ -19,12 +19,15 @@ import io.confluent.connect.hdfs.HdfsSinkConnectorConfig;
 import io.confluent.connect.storage.format.RecordWriter;
 import io.confluent.connect.storage.format.RecordWriterProvider;
 import io.confluent.connect.storage.hive.HiveSchemaConverter;
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hdfs.protocol.AlreadyBeingCreatedException;
 import org.apache.hadoop.hive.ql.io.orc.OrcFile;
 import org.apache.hadoop.hive.ql.io.orc.OrcStruct;
 import org.apache.hadoop.hive.ql.io.orc.Writer;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo;
+import org.apache.hadoop.ipc.RemoteException;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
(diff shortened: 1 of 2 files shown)


src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
```diff
@@ -99,6 +103,24 @@ public void preFooterWrite(OrcFile.WriterContext writerContext) {
...
+              /*
+               The current version of OrcFile.WriterOptions does not supports file overwriting
+               therefore deleting tmp file whenever`AlreadyBeingCreatedException` is thrown.
+                */
+              FileSystem fileSystem = FileSystem.get(
+                  URI.create(conf.getUrl()),
+                  conf.getHadoopConfiguration()
+              );
+              fileSystem.delete(path, true);
+            } catch (IOException ex) {
```
ilanjiR, line 117: Can we log this error message as well?

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
```diff
@@ -99,6 +103,24 @@ public void preFooterWrite(OrcFile.WriterContext writerContext) {
...
+            try {
+              /*
+               The current version of OrcFile.WriterOptions does not supports file overwriting
+               therefore deleting tmp file whenever`AlreadyBeingCreatedException` is thrown.
+                */
+              FileSystem fileSystem = FileSystem.get(
+                  URI.create(conf.getUrl()),
+                  conf.getHadoopConfiguration()
+              );
+              fileSystem.delete(path, true);
```
ilanjiR, line 116: Not too familiar with HDFS. Why are we deleting the file for ORC here? Why is `schema=null` not enough like with AVRO

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
SanchayGupta1197: Have added the exception to the log below.

src/main/java/io/confluent/connect/hdfs/orc/OrcRecordWriterProvider.java
SanchayGupta1197: The current version of `org.apache.hadoop.hive.ql.io.orc.OrcFile.WriterOptions` does not supports file overwriting therefore deleting tmp file whenever`AlreadyBeingCreatedException` is thrown.

PR #532, opened 2020-12-08
Fix the data loss problem

lightzhao:
## Problem
When HDFS fails, the data being written will be lost.

## Solution
 it is necessary to record the offset. When HDFS is recovered, the offset will be reset and rewind to ensure that the data is not lost.

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -406,6 +406,7 @@ public class TopicPartitionWriter {
         throw new RuntimeException(e);
       } catch (ConnectException e) {
         log.error("Exception on topic partition {}: ", tp, e);
+        context.offset(tp, offset);
         failureTime = time.milliseconds();
         setRetryTimeout(timeoutMs);
         break;


lightzhao:
@Ewen Cheslack-Postava. can you help me review?

lightzhao:
@DimaD can you help me review?

lightzhao:
@levzem hello， can you help me review ? thanks.

levzem:
@lightzhao first of all, the build is failing and second of all, HDFS uses the WAL and the offsets in the committed HDFS files to recover and keep track of the offset as you can see [here](https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L266). In fact, resetting the offset to something the connector might not have committed yet can result in data loss.

HDFS can strict EOS guarantee and if we are experiencing data loss that is a serious bug. If you have concrete evidence of this happening, please post that in the description of the PR, otherwise this change is not necessary.

lightzhao:
> @lightzhao first of all, the build is failing and second of all, HDFS uses the WAL and the offsets in the committed HDFS files to recover and keep track of the offset as you can see [here](https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L266). In fact, resetting the offset to something the connector might not have committed yet can result in data loss.
> 
> HDFS can strict EOS guarantee and if we are experiencing data loss that is a serious bug. If you have concrete evidence of this happening, please post that in the description of the PR, otherwise this change is not necessary.

@levzem First of all, thank you very much. This is a data loss problem encountered online. At that time, NAMENODE failed, but the task did not exit. The data during the 1-minute period that caused the namenode failure was lost. I can reproduce it in the test environment. The recover() method mentioned is to obtain the offset from the HDFS file when the task starts, but if the task does not exit and restart, the recover() method is not called, and the data may be lost. Is there a problem with my understanding?

lightzhao:
If NAMENODE fails, record the offset, when NAMENODE is restored, the offset can be played back, so at least the data will not be lost

PR #534, opened 2020-12-09
CC-13111: fix Kerberos tests interfering with non-Kerberos tests

levzem:

## Problem
when running all the tests together in Intellij the Kerberos tests interfere with the non-Kerberos tests and cause them fail because the UserGroupInformation is set by the connector and that is a static instance

## Solution
reset the UGI between each test
bonus: refactored the tests to take 30 seconds down from 7 minutes by reusing a HDFS instance for each test

- [x] yes
- [ ] no

##### If yes, where?
HDFS3 has the same issue

## Test Strategy
run all previous tests and make sure they are green when run simultaneously 

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
`10.0.x` is earliest because `reset()` isn't provided in earlier Hadoop versions

### DIFF
--- a/src/test/java/io/confluent/connect/hdfs/TestWithMiniDFSCluster.java
+++ b/src/test/java/io/confluent/connect/hdfs/TestWithMiniDFSCluster.java
@@ -42,6 +42,8 @@ import java.util.Set;
 import io.confluent.connect.hdfs.filter.TopicPartitionCommittedFileFilter;
 import io.confluent.connect.hdfs.partitioner.Partitioner;
 import io.confluent.connect.storage.common.StorageCommonConfig;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


src/test/java/io/confluent/connect/hdfs/TestWithSecureMiniDFSCluster.java
```diff
@@ -53,18 +54,36 @@
...
+
+  @AfterClass
+  public static void cleanup() throws IOException {
+    if (fs != null) {
+      fs.close();
+    }
+    if (cluster != null) {
+      cluster.shutdown(true);
+    }
+    UserGroupInformation.reset();
```
dosvath, line 82: We should reset the UGI after the non-kerberos tests as well, that way we don't have to rely on the order `kerb-tests -> regular tests`.

dosvath (commented):
Looks good, one comment and pending green build.

src/test/java/io/confluent/connect/hdfs/TestWithMiniDFSCluster.java
```diff
@@ -42,22 +42,41 @@
...
+  }
+
+  @AfterClass
+  public static void cleanup() throws IOException {
+    if (fs != null) {
+      fs.close();
+    }
+    if (cluster != null) {
+      cluster.shutdown(true);
+    }
```
dosvath, line 77:
```suggestion
    }
    UserGroupInformation.reset();
```

dosvath: To be consistent with https://github.com/confluentinc/kafka-connect-hdfs3/pull/44/files

levzem: there is no need for this because it is run after the kerberos tests and cleans everything up

dosvath (approved):
One comment, LGTM otherwise.

dosvath (approved):

PR #540, opened 2021-01-19
MINOR: Add task.id and additional logging when requesting a lease

dosvath:
- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -303,7 +303,7 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
   private final Map<String, ComposableConfig> propertyToConfig = new HashMap<>();
   private final Set<AbstractConfig> allConfigs = new HashSet<>();
   private Configuration hadoopConfig;
-  private String taskId;
+  private int taskId;
 
   public HdfsSinkConnectorConfig(Map<String, String> props) {
     this(newConfigDef() , addDefaults(props));
@@ -318,7 +318,8 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
     partitionerConfig = new PartitionerConfig(partitionerConfigDef, originalsStrings());
     this.name = parseName(originalsStrings());
     this.hadoopConfig = new Configuration();
-    taskId = props.get(TASK_ID_CONFIG_NAME);
+    taskId = props.get(TASK_ID_CONFIG_NAME) != null ?
+        Integer.parseInt(props.get(TASK_ID_CONFIG_NAME)) : -1;
     addToGlobal(hiveConfig);
     addToGlobal(partitionerConfig);
     addToGlobal(commonConfig);
@@ -499,7 +500,7 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
   }
 
   public int getTaskId() {
-    return taskId != null ? Integer.parseInt(taskId) : -1;
+    return taskId;
   }
 
   public static void main(String[] args) {
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
@@ -318,8 +318,8 @@ public class HdfsSinkConnectorConfig extends StorageSinkConnectorConfig {
     partitionerConfig = new PartitionerConfig(partitionerConfigDef, originalsStrings());
     this.name = parseName(originalsStrings());
     this.hadoopConfig = new Configuration();
-    taskId = props.get(TASK_ID_CONFIG_NAME) != null ?
-        Integer.parseInt(props.get(TASK_ID_CONFIG_NAME)) : -1;
+    taskId = props.get(TASK_ID_CONFIG_NAME) != null
+        ? Integer.parseInt(props.get(TASK_ID_CONFIG_NAME)) : -1;
     addToGlobal(hiveConfig);
     addToGlobal(partitionerConfig);
     addToGlobal(commonConfig);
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -151,7 +151,7 @@ public void truncate() throws ConnectException {
 
   @Override
   public void close() throws ConnectException {
```
xiangxin72, line 153: is there any where in the plugin code that this close() is supposed to be called? would it be useful to also add some logging there?

dosvath:
@avocader I think this is the line you wanted to get the ID on right? https://github.com/confluentinc/kafka-connect-hdfs/blob/60184b5a33d0058a42b5c559fc3feaf9c0b6dda1/src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java#L76

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: It's called by close in the TPW (https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java)

dosvath: What logging did you have in mind?

xiangxin72: as long as you think there are sufficient log leading up to this part then that should be fine. I guess my only worry was, it's possible the code somehow never reached to this point, then this line won't show up at all and we'd still need some info to help us trace back to where we think should be invoked. Maybe DEBUG level should be giving us all those info?

avocader:
> @avocader I think this is the line you wanted to get the ID on right?
> 
> https://github.com/confluentinc/kafka-connect-hdfs/blob/60184b5a33d0058a42b5c559fc3feaf9c0b6dda1/src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java#L76

I think so. Also, as we discussed, we can generate some unique identifier(`UUID.randomUUID();`) on the acquisition and when the file is being closed.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: gotcha, yes I will add some on higher level

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -154,13 +154,15 @@ public void open(Collection<TopicPartition> partitions) {
 
   @Override
   public void close(Collection<TopicPartition> partitions) {
+    log.debug("Closing HDFS Sink Task");
```
avocader, line 157: Should we also add a log statement to the `open()` method for parity?

xiangxin72 (approved):
thanks!

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -73,7 +76,13 @@ public void acquireLease() throws ConnectException {
...
+          log.info(
+              "Successfully acquired lease, {}-{}, FSWAL UUID {}, File {}",
+              conf.getName(),
+              conf.getTaskId(),
+              logFileId,
+              logFile
+          );
         }
         break;
       } catch (RemoteException e) {
```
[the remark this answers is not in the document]
kkonstantine: And even the exception messages as well

kkonstantine (commented):
Looks good. A few minor comments

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -151,7 +160,13 @@ public void truncate() throws ConnectException {
 
   @Override
   public void close() throws ConnectException {
-    log.debug("Closing WAL");
```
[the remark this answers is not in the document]
dosvath: yes info should be good here

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -42,13 +43,15 @@
   private String logFile = null;
   private HdfsSinkConnectorConfig conf = null;
   private HdfsStorage storage = null;
+  private UUID logFileId;
```
kkonstantine, line 46: this now seems unused

kkonstantine: I think my comment remains. Why do we need a UUID?

dosvath: In some recent logs we saw that a WAL lease was created by one UUID but closed by 3 different UUIDs + the original UUID, it would help us investigate if any potential issues arise.

dosvath: A second use would be for the case if two connectors with the same name are started, then the UUID would be the only differentiator in those logs.

dosvath: We could also opt for a more compact random number if that's preferable

kkonstantine: The problem with a random part is that we can't map it back to a specific task. 

I don't there's a strong case for adding it. The task ID should suffice to identify the task that issues the log messages. 
Finally, the use case of running multiple connect clusters with the connectors that have the same name across cluster is fairly uncommon to support it here IMO.

dosvath: Updated PR

dosvath:
@kkonstantine ready for another pass

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -495,6 +498,10 @@ private static void addAllConfigKeys(ConfigDef container, ConfigDef other, Set<S
     }
   }
 
+  public int getTaskId() {
+    return taskId != null ? Integer.parseInt(taskId) : -1;
```
kkonstantine, line 502: can we declare this member `final` and move this logic in the constructor now? That'd be preferable if we can

kkonstantine (approved):
One last comment I think. 
But otherwise, LGTM

PR #541, opened 2021-01-20
CCMSG-768: fix concurrent modification exception when closing TPW

levzem:

## Problem
```
[2021-01-13 11:20:56,910] ERROR WorkerSinkTask{id=KAFKA-HDP-NGBI.AUDIENCIAS-MDIC-2} Task threw an uncaught and unrecoverable exception (org.apache.kafka.connect.runtime.WorkerTask)
java.util.ConcurrentModificationException
	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1445)
	at java.util.HashMap$KeyIterator.next(HashMap.java:1469)
	at io.confluent.connect.hdfs.TopicPartitionWriter.close(TopicPartitionWriter.java:462)
	at io.confluent.connect.hdfs.DataWriter.close(DataWriter.java:471)
	at io.confluent.connect.hdfs.HdfsSinkTask.close(HdfsSinkTask.java:158)
```

which is thrown here in [TPW::close](https://github.com/confluentinc/kafka-connect-hdfs/blob/v10.0.0/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L462)

because [TPW::closeTempFile](https://github.com/confluentinc/kafka-connect-hdfs/blob/v10.0.0/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L739) removes the writer

the [wal lease is closed after this process](https://github.com/confluentinc/kafka-connect-hdfs/blob/v10.0.0/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L498)

therefore if the connector fails in this step, it does not close the WAL, and therefore does not release its lease properly

## Solution
the solution is simple, just use `tempFiles.keySet()` instead because it is not cleared
https://github.com/confluentinc/kafka-connect-hdfs/blob/v10.0.0/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L747

- [ ] yes
- [x] no

## Test Strategy
added a test that makes sure that close work appropriately when many writers are open
before this change, the unit test would fail

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests

## Release Plan

 
`5.0.x` where this code was introduced


dosvath (approved):
LGTM, thanks for the tests @levzem.

kkonstantine:
@levzem @dosvath is it correct to assume that tempFiles is a collection that only grows in size? After a tempFile is removed the collection does not remove the key (encoded partition). Such an issue might become visible in long running tasks and when the encoded partitions progress frequently. But is a bug currently. Probably good to address it in a separate fix.

levzem:
@kkonstantine it depends on what the partitioner is because the encoded partition will be a finite list for the `DefaultPartitioner` but ever growing with any of the `TimePartitioners`

kkonstantine:
@levzem we can assume that time based partitioners are the majority.

PR #553, opened 2021-03-17
Support N topics written under the same directory

JozoVilcek:
## Problem

When using a custom partitioner, one can opt out from having topic data separated into topic specific directory. But when writing data from N topics under the same directory, the recovery of offsets is looking for files under paths which does not exists. 

## Solution

Detect if partitioner us adding topic name to the output path. If not, exclude it from the search path for committed files -> start search from parent directory. This however does not cover generic case when partitioner chooses some exotic path strategy.

- [ ] yes
- [x] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [x] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -399,7 +399,8 @@ public class DataWriter {
 
     try {
       for (String topic : topics) {
-        String topicDir = FileUtils.topicDirectory(url, topicDirs.get(topic), topic);
+        String topicDir = FileUtils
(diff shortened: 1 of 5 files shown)


JozoVilcek:
[clabot:check]

JozoVilcek:
Does this PR make sense? Any kind of feedback is much appreciated

JozoVilcek:
@kpatelatwork I did find you active in other MRs. Can you help me how can I bring contributions into the attention of code owners? I have problem to get feedback on this and few other old MRs

JozoVilcek:
@janjwerner-confluent , @rishabhbits038 could you please review this request and let me know the feedback if this has a chance to get in. If yes I would work on updating the PR and make it green

PR #556, opened 2021-04-13
CCMSG-670: Use WAL file as primary for offset recovery

dosvath:
## Problem
The HDFS connector tracks Kafka offsets through the names of the files written to HDFS. 
1. On startup, the connector attempts to recover the most recently committed offsets by recursively scanning the `topics.dir` for each topic partition it has been assigned, and then resumes consuming from those offsets. 

2. On write, when the schema is initialized the connector performs the same recursive search. 

The recursive file listing operation becomes expensive when there are thousands or millions of files and the connector takes a long time to start. 

## Solution

Use the stored offsets from the `WAL` file as a primary method for recovery. Fall-back on the current approach as a backup and for backwards compatibility when the `WAL` is not readable or non-existent. 

**Summary of functional changes**

`TopicPartitionWriter.java`
* The offsets are now attempted to be reset before truncating the existing `WAL`. Currently, truncating before resetting would cause the old `WAL` to be deleted before extracting the offset information. 
* Reset the offsets based on the `WAL` and if not applicable use the current method as fallback. 
* Use the file path of the latest offsets file that was extracted from the `WAL` for the schema initialization as well to avoid the recursive search. Fallback on recursive. 

`FSWAL.java`
* Added offset extraction logic for both recent and truncated (old) `WAL`. 

`WALFile.java`
* Added method to seek to the beginning of the first record for a file `Reader`.

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [X] Manual tests

## Release Plan

 
`10.0.x` for major change.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -117,7 +117,8 @@ public class HdfsSinkTask extends SinkTask {
 
     log.info("The connector relies on offsets in HDFS filenames, but does commit these offsets to "
         + "Connect to enable monitoring progress of the HDFS connector. Upon startup, the HDFS "
-        + "Connector restores offsets from filenames in HDFS. In the absence of files in HDFS, "
+        + "Connector restores offsets from the WAL log files, if these are not present it "
+        + "uses the filenames in HDFS. In the absence of files in HDFS, "
         + "the connector will attempt to find offsets for its consumer group in the "
         + "'__consumer_offsets' topic. If offsets are not found, the consumer will "
         + "rely on the reset policy specified in the 'consumer.auto.offset.reset' property to "
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -115,8 +115,9 @@ public class HdfsSinkTask extends SinkTask {
       throw e;
     }
 
-    log.info("The connector relies on offsets in HDFS filenames, but does commit these offsets to "
-        + "Connect to enable monitoring progress of the HDFS connector. Upon startup, the HDFS "
+    log.info("The connector relies on offsets in the WAL files, if these are not present it uses "
+        + "the filenames in HDFS, and in both cases commits these offsets to Connect to "
+        + "enable monitoring progress of the HDFS connector. Upon startup, the HDFS "
         + "Connector restores offsets from the WAL log files, if these are not present it "
         + "uses the filenames in HDFS. In the absence of files in HDFS, "
         + "the connector will attempt to find offsets for its consumer group in the "
--- a/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
+++ b/src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
@@ -116,7 +116,7 @@ public class HdfsSinkTask extends SinkTask {
     }
 
     log.info("The connector relies on offsets in the WAL files, if these are not present it uses "
-        + "the filenames in HDFS, and in both cases commits these offsets to Connect to "
+        + "the filenames in HDFS. In both cases the connector commits offsets to Connect to "
         + "enable monitoring progress of the HDFS connector. Upon startup, the HDFS "
         + "Connector restores offsets from the WAL log files, if these are not present it "
         + "uses the filenames in HDFS. In the absence of files in HDFS, "
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -16,6 +16,7 @@
 package io.confluent.connect.hdfs;
 
 import io.confluent.connect.hdfs.wal.FSWAL;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.Path;
 import org.apache.kafka.common.TopicPartition;
@@ -99,6 +100,7 @@ public class TopicPartitionWriter {
   private final AvroData avroData;
   private final Set<String> appended;
   private long offset;
+  private String latestOffsetsFile;
   private final Map<String, Long> startOffsets;
   private final Map<String, Long> offsets;
   private final long timeoutMs;
@@ -205,6 +207,7 @@ public class TopicPartitionWriter {
     tempFiles = new HashMap<>();
     appended = new HashSet<>();
     startOffsets = new HashMap<>();
+    latestOffsetsFile = null;
     offsets = new HashMap<>();
     state = State.RECOVERY_STARTED;
     failureTime = -1L;
@@ -338,24 +341,7 @@ public class TopicPartitionWriter {
             nextState();
           case WRITE_PARTITION_PAUSED:
             if (currentSchema == null) {
-              log.debug("Current schema is null, getting schema from file with latest offsets.");
-              if (compatibility != StorageSchemaCompatibility.NONE && offset != -1) {
-                String topicDir = FileUtils.topicDirectory(url, topicsDir, tp.topic());
-                CommittedFileFilter filter = new TopicPartitionCommittedFileFilter(tp);
-                log.trace("Fetching file with max offset.");
-                FileStatus fileStatusWithMaxOffset = FileUtils.fileStatusWithMaxOffset(
-                    storage,
-                    new Path(topicDir),
-                    filter
-                );
-                if (fileStatusWithMaxOffset != null) {
-                  currentSchema = schemaFileReader.getSchema(
-                      connectorConfig,
-                      fileStatusWithMaxOffset.getPath()
-                  );
-                  log.trace("Retrieved schema from file with latest offset.");
-                }
-              }
+              fetchSchemaFromLatestFile();
             }
             SinkRecord record = buffer.peek();
             currentRecord = record;
@@ -629,13 +615,15 @@ public class TopicPartitionWriter {
   private boolean readOffsetFromWAL() {
     // MemoryWAL is used for testing
     if (wal instanceof FSWAL) {
-      long lastCommittedOffsetToHdfs = ((FSWAL) wal).extractLatestOffsetFromWAL();
-      if (lastCommittedOffsetToHdfs == -1) {
+      Pair<Long, String> latestOffsetEntry = ((FSWAL) wal).extractLatestOffsetFromWAL();
+      if (latestOffsetEntry == null) {
         return false;
       }
-      log.trace("Last committed offset based on WAL: {}", lastCommittedOffsetToHdfs);
-      offset = lastCommittedOffsetToHdfs + 1;
+      log.trace("Last committed offset based on WAL: {}", latestOffsetEntry.getKey());
+      offset = latestOffsetEntry.getKey() + 1;
       log.trace("Next offset to read: {}", offset);
+      latestOffsetsFile = latestOffsetEntry.getValue();
+      log.trace("Set latest offsets file {}", latestOffsetsFile);
       return true;
     }
     return false;
@@ -708,9 +696,7 @@ public class TopicPartitionWriter {
   }
 
   private void truncateWAL() throws ConnectException {
-    if (!recovered) {
-      wal.truncate();
-    }
+    wal.truncate();
   }
 
   private void resetOffsets() throws ConnectException {
@@ -809,6 +795,37 @@ public class TopicPartitionWriter {
 
       recordCounter = 0;
       throw connectException;
+    }
+  }
+
+  private void fetchSchemaFromLatestFile() {
+    if (compatibility != StorageSchemaCompatibility.NONE && offset != -1) {
+      log.debug("Current schema is null, getting schema from file with latest offsets.");
+      // check on WAL restored latest file to avoid recursive search
+      if (recovered && latestOffsetsFile != null) {
+        currentSchema = schemaFileReader.getSchema(
+            connectorConfig,
+            new Path(latestOffsetsFile)
+        );
+        log.trace("Retrieved schema from file with latest offset {}", latestOffsetsFile);
+      } else {
+        // if WAL restored file is N/A fallback on existing recursive scan approach
+        String topicDir = FileUtils.topicDirectory(url, topicsDir, tp.topic());
+        CommittedFileFilter filter = new TopicPartitionCommittedFileFilter(tp);
+        log.trace("Fetching file with max offset.");
+        FileStatus fileStatusWithMaxOffset = FileUtils.fileStatusWithMaxOffset(
+            storage,
+            new Path(topicDir),
+            filter
+        );
+        if (fileStatusWithMaxOffset != null) {
+          currentSchema = schemaFileReader.getSchema(
+              connectorConfig,
+              fileStatusWithMaxOffset.getPath()
+          );
+          log.trace("Retrieved schema from file with latest offset {}", latestOffsetsFile);
+        }
+      }
     }
   }
 
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -15,6 +15,7 @@
 
 package io.confluent.connect.hdfs;
 
+import io.confluent.connect.hdfs.wal.FSWAL;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.Path;
 import org.apache.kafka.common.TopicPartition;
@@ -258,14 +259,14 @@ public class TopicPartitionWriter {
           applyWAL();
           nextState();
         case WAL_APPLIED:
-          log.debug("Start recovery state: Truncate WAL for topic partition {}", tp);
-          truncateWAL();
-          nextState();
-        case WAL_TRUNCATED:
           log.debug("Start recovery state: Reset Offsets for topic partition {}", tp);
           resetOffsets();
           nextState();
         case OFFSET_RESET:
+          log.debug("Start recovery state: Truncate WAL for topic partition {}", tp);
+          truncateWAL();
+          nextState();
+        case WAL_TRUNCATED:
           log.debug("Start recovery state: Resume for topic partition {}", tp);
           resume();
           nextState();
@@ -337,9 +338,11 @@ public class TopicPartitionWriter {
             nextState();
           case WRITE_PARTITION_PAUSED:
             if (currentSchema == null) {
+              log.debug("Current schema is null, getting schema from file with latest offsets.");
               if (compatibility != StorageSchemaCompatibility.NONE && offset != -1) {
                 String topicDir = FileUtils.topicDirectory(url, topicsDir, tp.topic());
                 CommittedFileFilter filter = new TopicPartitionCommittedFileFilter(tp);
+                log.trace("Fetching file with max offset.");
                 FileStatus fileStatusWithMaxOffset = FileUtils.fileStatusWithMaxOffset(
                     storage,
                     new Path(topicDir),
@@ -350,6 +353,7 @@ public class TopicPartitionWriter {
                       connectorConfig,
                       fileStatusWithMaxOffset.getPath()
                   );
+                  log.trace("Retrieved schema from file with latest offset.");
                 }
               }
             }
@@ -599,7 +603,7 @@ public class TopicPartitionWriter {
     return periodicRotation || scheduledRotation || messageSizeRotation;
   }
 
-  private void readOffset() throws ConnectException {
+  private void readOffsetFromFilenames() throws ConnectException {
     String path = FileUtils.topicDirectory(url, topicsDir, tp.topic());
     CommittedFileFilter filter = new TopicPartitionCommittedFileFilter(tp);
     FileStatus fileStatusWithMaxOffset = FileUtils.fileStatusWithMaxOffset(
@@ -615,6 +619,26 @@ public class TopicPartitionWriter {
       offset = lastCommittedOffsetToHdfs + 1;
       log.trace("Next offset to read: {}", offset);
     }
+  }
+
+  /**
+   * Read the most recently committed offset from the WAL file.
+   *
+   * @return whether a valid offset was read
+   */
+  private boolean readOffsetFromWAL() {
+    // MemoryWAL is used for testing
+    if (wal instanceof FSWAL) {
+      long lastCommittedOffsetToHdfs = ((FSWAL) wal).extractLatestOffsetFromWAL();
+      if (lastCommittedOffsetToHdfs == -1) {
+        return false;
+      }
+      log.trace("Last committed offset based on WAL: {}", lastCommittedOffsetToHdfs);
+      offset = lastCommittedOffsetToHdfs + 1;
+      log.trace("Next offset to read: {}", offset);
+      return true;
+    }
+    return false;
   }
 
   private void pause() {
@@ -691,7 +715,10 @@ public class TopicPartitionWriter {
 
   private void resetOffsets() throws ConnectException {
     if (!recovered) {
-      readOffset();
+      if (!readOffsetFromWAL()) {
+        // fallback on existing approach
+        readOffsetFromFilenames();
+      }
       // Note that we must *always* request that we seek to an offset here. Currently the
       // framework will still commit Kafka offsets even though we track our own (see KAFKA-3462),
       // which can result in accidentally using that offset if one was committed but no files
@@ -927,8 +954,8 @@ public class TopicPartitionWriter {
     RECOVERY_STARTED,
     RECOVERY_PARTITION_PAUSED,
     WAL_APPLIED,
-    WAL_TRUNCATED,
     OFFSET_RESET,
+    WAL_TRUNCATED,
     WRITE_STARTED,
     WRITE_PARTITION_PAUSED,
     SHOULD_ROTATE,
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -617,7 +617,7 @@ public class TopicPartitionWriter {
 
     // Use the recursive filename scan approach
     log.debug("Could not use WAL approach for recovering offsets, "
-        + "searching for latest offsets on filesystem.");
+        + "searching for latest offsets on HDFS.");
     String path = FileUtils.topicDirectory(url, topicsDir, tp.topic());
     CommittedFileFilter filter = new TopicPartitionCommittedFileFilter(tp);
     FileStatus fileStatusWithMaxOffset = FileUtils.fileStatusWithMaxOffset(
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -600,7 +600,24 @@ public class TopicPartitionWriter {
     return periodicRotation || scheduledRotation || messageSizeRotation;
   }
 
-  private void readOffsetFromFilenames() throws ConnectException {
+  /**
+   * Read the offset of most recent record in HDFS.
+   * Attempt to read the offset from the WAL file and fall-back on a recursive search of filenames.
+   */
+  private void readOffset() {
+    // Use the WAL file to attempt to extract the recent offsets
+    FilePathOffset latestOffsetEntry = wal.extractLatestOffset();
+    if (latestOffsetEntry != null) {
+      long lastCommittedOffset = latestOffsetEntry.getOffset();
+      log.trace("Last committed offset based on WAL: {}", lastCommittedOffset);
+      offset = lastCommittedOffset + 1;
+      log.trace("Next offset to read: {}", offset);
+      return;
+    }
+
+    // Use the recursive filename scan approach
+    log.debug("Could not use WAL approach for recovering offsets, "
+        + "searching for latest offsets on filesystem.");
     String path = FileUtils.topicDirectory(url, topicsDir, tp.topic());
     CommittedFileFilter filter = new TopicPartitionCommittedFileFilter(tp);
     FileStatus fileStatusWithMaxOffset = FileUtils.fileStatusWithMaxOffset(
@@ -616,23 +633,6 @@ public class TopicPartitionWriter {
       offset = lastCommittedOffsetToHdfs + 1;
       log.trace("Next offset to read: {}", offset);
     }
-  }
-
-  /**
-   * Read the most recently committed offset from the WAL file.
-   *
-   * @return whether a valid offset was read
-   */
-  private boolean readOffsetFromWAL() {
-    FilePathOffset latestOffsetEntry = wal.extractLatestOffset();
-    if (latestOffsetEntry == null) {
-      return false;
-    }
-    long lastCommittedOffset = latestOffsetEntry.getOffset();
-    log.trace("Last committed offset based on WAL: {}", lastCommittedOffset);
-    offset = lastCommittedOffset + 1;
-    log.trace("Next offset to read: {}", offset);
-    return true;
   }
 
   private void pause() {
@@ -707,12 +707,7 @@ public class TopicPartitionWriter {
 
   private void resetOffsets() throws ConnectException {
     if (!recovered) {
-      if (!readOffsetFromWAL()) {
-        // fallback on recursive filename search approach
-        log.debug("Could not use WAL approach for recovering offsets, "
-            + "searching for latest offsets on filesystem.");
-        readOffsetFromFilenames();
-      }
+      readOffset();
       // Note that we must *always* request that we seek to an offset here. Currently the
       // framework will still commit Kafka offsets even though we track our own (see KAFKA-3462),
       // which can result in accidentally using that offset if one was committed but no files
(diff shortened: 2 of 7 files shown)


src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +188,138 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
     }
   }
 
+
```
kkonstantine, line 191: nit: extra line

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +188,138 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
-      storage.delete(oldLogFile);
-      storage.commit(logFile, oldLogFile);
+      String oldLogFile = logFile + OLD_LOG_EXTENSION;
+      if (storage.exists(oldLogFile)) {
+        storage.delete(oldLogFile);
+      }
+      if (storage.exists(logFile)) {
+        storage.commit(logFile, oldLogFile);
+      } else {
+        log.trace("Attempted to truncate WAL but current WAL cannot be found {}", logFile);
```
kkonstantine, line 321: seems this might need to be `debug` given the log line at the beginning of this method

kkonstantine (commented):
Thanks @dosvath 
It'd be good to improve the startup time. I took a first pass and left a few comments. 
But I have to admit I'd like to understand better (and remember) the implications of this change and if there are any risks. 

Having said that, how come and we target 5.5.x ? Would branch 10.0.x be a better choice?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -15,6 +15,8 @@
 
 package io.confluent.connect.hdfs;
 
+import io.confluent.connect.hdfs.wal.FSWAL;
+import org.apache.commons.lang3.tuple.Pair;
```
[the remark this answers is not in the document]
dosvath: I tried using the `Pair` class from the package `javafx.util.Pair` with a [dockerized](https://github.com/vdesabou/kafka-docker-playground/tree/master/connect/connect-hdfs2-sink) testing setup and that class failed due to being Java version dependent. Do you have any alternate package suggestions?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -617,6 +607,28 @@ private void readOffset() throws ConnectException {
...
   }
 
+  /**
+   * Read the most recently committed offset from the WAL file.
+   *
+   * @return whether a valid offset was read
+   */
+  private boolean readOffsetFromWAL() {
+    // MemoryWAL is used for testing
+    if (wal instanceof FSWAL) {
```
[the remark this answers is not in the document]
dosvath: We would need to add onto the [interface in storage common](https://github.com/confluentinc/kafka-connect-storage-common/blob/4cce85d8f4106ea0488738b4ea08ff57fafaff28/wal/src/main/java/io/confluent/connect/storage/wal/WAL.java#L20) I can follow up there if that's preferable.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -617,6 +607,28 @@ private void readOffset() throws ConnectException {
...
+      Pair<Long, String> latestOffsetEntry = ((FSWAL) wal).extractLatestOffsetFromWAL();
+      if (latestOffsetEntry == null) {
+        return false;
+      }
+      log.trace("Last committed offset based on WAL: {}", latestOffsetEntry.getKey());
```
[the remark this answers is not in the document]
dosvath: Do you mean we should avoid a duplicate call on the next line 623? We can assign the result to a variable to use it in both of these subsequent lines.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -684,14 +696,17 @@ private void applyWAL() throws ConnectException {
   }
 
   private void truncateWAL() throws ConnectException {
-    if (!recovered) {
```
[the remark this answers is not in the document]
dosvath: The previous logic called truncate before recovering the offsets. With the new changes the [offsets are recovered before](https://github.com/confluentinc/kafka-connect-hdfs/blob/a5865bd03872420599c04e76a6985b8fe15f14d2/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L266) the attempt to truncate so `recover` will not have the chance to be false by the time this is reached.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -785,6 +800,39 @@ private void closeTempFile() {
     }
   }
 
+  private void fetchSchemaFromLatestFile() {
+    if (compatibility != StorageSchemaCompatibility.NONE && offset != -1) {
+      log.debug("Current schema is null, getting schema from file with latest offsets.");
```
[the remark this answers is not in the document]
dosvath: The schema has not been set, will update with `not set` instead of `null`. Yes we can add the partition information.

src/main/java/io/confluent/connect/hdfs/HdfsSinkTask.java
```diff
@@ -115,9 +115,11 @@ public void start(Map<String, String> props) {
       throw e;
     }
 
-    log.info("The connector relies on offsets in HDFS filenames, but does commit these offsets to "
-        + "Connect to enable monitoring progress of the HDFS connector. Upon startup, the HDFS "
-        + "Connector restores offsets from filenames in HDFS. In the absence of files in HDFS, "
+    log.info("The connector relies on offsets in the WAL files, if these are not present it uses "
```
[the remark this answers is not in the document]
dosvath: Agree, it will read better.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +188,138 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+  /**
+   * Extract the last filled BEGIN-END block of entries from the WAL.
+   *
+   * @param reader the WAL file reader
+   * @return the last batch of entries, may be empty if the WAL
+   *         is empty or only contains empty BEGIN-END blocks
+   * @throws IOException error on reading the WAL file
+   */
+  private ArrayList<String> getLastFilledBlockFromWAL(Reader reader) throws IOException {
+    //temp filenames don't contain offset info, use committed only (WAL entry values)
```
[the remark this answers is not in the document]
dosvath: added detailed comment

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -684,14 +696,17 @@ private void applyWAL() throws ConnectException {
...
-      wal.truncate();
-    }
+    wal.truncate();
   }
 
   private void resetOffsets() throws ConnectException {
     if (!recovered) {
-      readOffset();
+      if (!readOffsetFromWAL()) {
+        // fallback on existing approach
```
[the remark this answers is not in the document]
dosvath: Updated detailed comment.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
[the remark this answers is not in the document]
kkonstantine: Don't use Pair. Define your own class. Java doesn't have definitions for pair and tuples unlike elsewhere. That's fine. 
(You might see on the web examples that use `AbstractMap.SimpleEntry<K,V>`. Don't 😄 )

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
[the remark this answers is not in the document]
dosvath: Note that `truncateWAL()` is only called in [recover()](https://github.com/confluentinc/kafka-connect-hdfs/blob/a5865bd03872420599c04e76a6985b8fe15f14d2/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L253), which means the recovery will always happen before the call. `wal.truncate()` will be safe to call here without the check since it is always called after [apply()](https://github.com/confluentinc/kafka-connect-hdfs/blob/a5865bd03872420599c04e76a6985b8fe15f14d2/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L262), and this PR also adds file existence checks in the FSWAL class.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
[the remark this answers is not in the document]
kkonstantine: The point of a log line being in trace is that it's not printed often. But we need to keep in mind that even though the log string might not get fully constructed, any arguments that need to be computed will be computed regardless of log level. Here this might not be very expensive and worth changing but I wanted to highlight that we need to be aware of the cost of trace messages. For example, what's the cost of the `latestOffsetEntry.getKey()` call? For expensive cases you'd need to use `if(isTraceEnabled())`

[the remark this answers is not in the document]
dosvath: I see, yes that is a good point to keep in mind. We don't want unnecessary computations.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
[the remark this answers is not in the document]
dosvath: @kkonstantine I updated the storage common repo https://github.com/confluentinc/kafka-connect-storage-common/pull/182

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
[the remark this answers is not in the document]
dosvath: Sounds good, will be updated once https://github.com/confluentinc/kafka-connect-storage-common/pull/182 is in

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +189,141 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+          reader.seekToFirstRecord();
+        }
+        List<String> committedFileBatch = getLastFilledBlockFromWAL(reader);
+        // At this point the committedFilenames list will contain the
+        // filenames in the last BEGIN-END block of the file. Find the latest offsets among these.
+        latestOffset = getLatestOffsetFromList(committedFileBatch);
+      }
+
+      // attempt to use old log file if recent WAL is empty or non-existent
+      if (latestOffset == null && storage.exists(oldWALFile)) {
```
erdody, line 221: Does it make sense to attempt this part if the above part throws IOException?

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +189,141 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+        log.trace("Could not find offset in log file {}. Using {} instead", logFile, oldWALFile);
+        Reader oldFileReader =
```
erdody, line 223: Try-with-resource here will make sure the reader is closed even after an exception.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +189,141 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+            new WALFile.Reader(conf.getHadoopConfiguration(), Reader.file(new Path(oldWALFile)));
+        List<String> committedFileBatch = getLastFilledBlockFromWAL(oldFileReader);
+        latestOffset = getLatestOffsetFromList(committedFileBatch);;
```
erdody, line 226: Nit: double semicolon.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +189,141 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+      }
+      return latestOffset;
+    } catch (IOException e) {
+      log.warn("Error restoring offsets from either {} or {} WAL files", logFile, oldWALFile);
+      return null;
+    }
+  }
+
+  /**
+   * Extract the last filled BEGIN-END block of entries from the WAL.
```
erdody, line 237: This would also return entries after the last BEGIN with no END, is that a concern? Should we throw an exception in that case?

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +189,141 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+   * @return the offset or -1 if not present
+   */
+  static long extractOffsetsFromFilePath(String fullPath) {
+    try {
+      if (fullPath != null) {
+        String latestFileName = Paths.get(fullPath).getFileName().toString();
+        return FileUtils.extractOffset(latestFileName);
+      }
+    } catch (IllegalArgumentException e) {
+      log.warn("Could not extract offsets from file path: {}", fullPath);
```
erdody, line 305: Can we also log the exception (or message)?

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: When we write a block of entries in [appendToWal() here](https://github.com/confluentinc/kafka-connect-hdfs/blob/c5073c7f831e377dbd1f63b9b08ccd40d6dcfb92/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L813) we always add the END right after. To be safe, we can definitely throw an exception.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -36,6 +41,7 @@
 public class FSWAL implements WAL {
 
   private static final Logger log = LoggerFactory.getLogger(FSWAL.class);
+  private static final String OLD_LOG_EXTENSION = ".1";
```
levzem, line 44:
```suggestion
  private static final String TRUNCATED_LOG_EXTENSION = ".1";
```

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -785,6 +801,40 @@ private void closeTempFile() {
...
+        log.debug("Could not find latest offsets file from WAL for schema initialization, "
+            + "searching for latest offsets file on filesystem.");
+        String topicDir = FileUtils.topicDirectory(url, topicsDir, tp.topic());
+        CommittedFileFilter filter = new TopicPartitionCommittedFileFilter(tp);
+        log.trace("Fetching file with max offset.");
+        FileStatus fileStatusWithMaxOffset = FileUtils.fileStatusWithMaxOffset(
+            storage,
+            new Path(topicDir),
+            filter
+        );
```
levzem, line 826: we already do this in `readOffsets()`, can we further optimize and only do this once, maybe save the return value of `readOffsets()`?

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +189,141 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
     try {
-      String oldLogFile = logFile + ".1";
-      storage.delete(oldLogFile);
-      storage.commit(logFile, oldLogFile);
+      String oldLogFile = logFile + OLD_LOG_EXTENSION;
+      if (storage.exists(oldLogFile)) {
+        storage.delete(oldLogFile);
+      }
+      if (storage.exists(logFile)) {
+        storage.commit(logFile, oldLogFile);
```
levzem, line 323: why do we need this?

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -785,6 +801,40 @@ private void closeTempFile() {
     }
   }
 
+  private void fetchSchemaFromLatestFile() {
+    if (compatibility != StorageSchemaCompatibility.NONE && offset != -1) {
+      log.debug("Current schema is not set for partition {}, "
+          + "getting schema from file with latest offsets.", tp);
+      // check on WAL restored latest file to avoid recursive search
+      if (recovered && latestOffsetsFile != null) {
+        currentSchema = schemaFileReader.getSchema(
```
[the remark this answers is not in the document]
dosvath: It would produce the same result as the approach with the max offset file, the `latestOffsetsFile` path that is passed to it is the same filepath that the max offsets file would have. This happens after `wal.apply()` so the filesystem and WAL file are in sync at this point.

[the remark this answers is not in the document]
dosvath: So falling back to the search method would also produce the same exception.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: It wouldn't for the following reason: the most recent WAL will have the most recent offsets, and falling back to the old WAL would cause to use an older offset (restore to an older point than the files which are in HDFS). If the recent `WAL` exists and throws an IOException the fallback should be the recursive file search, not the old `WAL`. Let me know if that makes sense.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
erdody: Thanks, I saw that, but I was thinking of something happening to that call before writing the end marker. Not sure if we catch that scenario earlier in the code. If there's not an obvious place where we're doing that, I would throw here.

erdody (approved):
LGTM, given that we incorporate the FSWAL updates. Thanks!

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: Thanks @erdody it is a good point, I revised the logic to handle the missing end marker accordingly. Take a look if you get a chance.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
dosvath: That is a good thought, I started attempting that when I realized the issue is the following:

Both `readoffsets()` and `fetchSchemaFromLatestFile()` are under the `write()` method. 

The `write()` method only calls `readOffsets()` in some cases, before the state is `WRITE_STARTED`. 
However, `write()` may call `fetchSchemaFromLatestFile()` on an execution when `readoffsets()` wasn't called. 

This means that the stored value for the max file status that we get from `readoffsets()` may be from an earlier (outdated) call to `write()`. In that case the max file needs to be recomputed here. 

This also made me realize it holds for the case of the readOffsetsFromWAL() as well. We can't rely on that saved file, so we'll need to extract from the WAL again here. I will update the PR.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +189,160 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+   *
+   * @param reader the WAL file reader
+   * @return the last batch of entries, may be empty if the WAL
+   *         is empty or only contains empty BEGIN-END blocks
+   * @throws IOException error on reading the WAL file
+   */
+  private List<String> getLastFilledBlockFromWAL(Reader reader) throws IOException {
+    // In a WAL entry the temp filenames (keys) don't contain offset info,
+    // so we only need to track the committed filename (values).
+    ArrayList<String> committedFilenames = new ArrayList<>();
```
kkonstantine, line 253: @dosvath in some cases a comment applies to several instances in the file. I think I forgot to mention to replace the rest of the impl declarations with the interface. Please check for `ArrayList` declarations and use `List`

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: Do you mean the `storage.commit(logFile, oldLogFile);` part? That is how the connector writes the recent `WAL` file to the truncated (old) file thereby creating the truncated file with the `.1` extension. If that commit wasn't present, the truncated `WAL`s wouldn't exist.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: Got it :)

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
dosvath: I updated the PR with the revision for this issue.

dosvath: On a second thought, the schema recovery logic should be followed up in a separate PR.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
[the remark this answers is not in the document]
levzem: looks like we remove the instance of now

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
levzem: sorry, i meant the `exists` calls

src/test/java/io/confluent/connect/hdfs/utils/MemoryWAL.java
```diff
@@ -15,14 +15,14 @@
 
 package io.confluent.connect.hdfs.utils;
 
+import io.confluent.connect.storage.wal.FilePathOffset;
```
levzem, line 18: nit import order

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -15,6 +15,11 @@
 
 package io.confluent.connect.hdfs.wal;
 
+import io.confluent.connect.storage.wal.FilePathOffset;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+
+import java.util.List;
```
levzem, line 22: nit imports

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -15,6 +15,8 @@
 
 package io.confluent.connect.hdfs;
 
+import io.confluent.connect.hdfs.wal.FSWAL;
+import io.confluent.connect.storage.wal.FilePathOffset;
```
levzem, line 19: nit imports

src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
```diff
@@ -72,4 +78,239 @@ public void testTruncatedVersionWalFileRecovery() throws Exception {
...
+  }
+
+  @Test
+  public void testOffsetsExtractedFromWALWithEmptyBlocks() throws Exception {
+    setUp();
+    String topicsDir = this.topicsDir.get(TOPIC_PARTITION.topic());
+    fs.delete(new Path(FileUtils.directoryName(url, topicsDir, TOPIC_PARTITION)), true);
+    HdfsStorage storage = new HdfsStorage(connectorConfig, url);
+    DataWriter hdfsWriter = new DataWriter(connectorConfig, context, avroData);
+    partitioner = hdfsWriter.getPartitioner();
```
levzem, line 108: might be able to use setup wal test here

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +189,166 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+        if (reader == null) {
+          reader = newWalFileReader();
+        } else {
+          // reset read position after apply()
+          reader.seekToFirstRecord();
+        }
+        List<String> committedFileBatch = getLastFilledBlockFromWAL(reader);
+        // At this point the committedFilenames list will contain the
+        // filenames in the last BEGIN-END block of the file. Find the latest offsets among these.
+        latestOffset = getLatestOffsetFromList(committedFileBatch);
```
levzem, line 218:
```suggestion
        return getLatestOffsetFromList(committedFileBatch);
```
then we can remove the null check in the if statement

levzem (commented):
overall looks solid,

are there existing tests for starting and stopping a task/recovering?

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: Here it would still be necessary to check on the result of this function. `getLatestOffsetFromList()` returns `null` when no offset is found. In that case we would need to move up the null check because we don't want to return `null` without checking the old file. In that case we may keep the current logic?

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: I added the exist calls due to seeing any exist checks in the chain of calls for the `delete` and `commit` file system methods.

dosvath:
> also would we consider this a bug fix or a new feature/behavior? because it should go to master if that is the case

Possibly, @kkonstantine any preference?

dosvath:
@levzem we have existing recovery tests for the task restart here in [testSinkTaskStartWithRecovery()](https://github.com/confluentinc/kafka-connect-hdfs/blob/d140586c87275283a5265913b4130a3fa16fd81f/src/test/java/io/confluent/connect/hdfs/HdfsSinkTaskTest.java#L202). These use the HDFS mini cluster. Additionally, in the `DataWriterAvroTest` the TPW itself is tested directly for the recovery here in [testRecovery()](https://github.com/confluentinc/kafka-connect-hdfs/blob/d140586c87275283a5265913b4130a3fa16fd81f/src/test/java/io/confluent/connect/hdfs/avro/DataWriterAvroTest.java#L88). The HDFS cluster is used for these as well.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
levzem: but its been working before? interesting

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,13 +188,166 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+      // attempt to use old log file if recent WAL is empty or non-existent
+      if (latestOffset == null && storage.exists(oldWALFile)) {
+        log.trace("Could not find offset in log file {}. Using {} instead", logFile, oldWALFile);
+        try (Reader oldFileReader =
+            new WALFile.Reader(conf.getHadoopConfiguration(), Reader.file(new Path(oldWALFile)))
+        ) {
+          List<String> committedFileBatch = getLastFilledBlockFromWAL(oldFileReader);
+          latestOffset = getLatestOffsetFromList(committedFileBatch);
+        }
+      }
```
levzem, line 229: this looks like some repeated code that we can factor out for readability i.e

`private FilePathOffset extractLatestOffset(String logFile);`

WDYT?

levzem (approved):
LGTM

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: Close, but the difference in initializing the old vs new readers needs to be kept, so not sure if it would be worth it for the two other lines (226, 227). For the recent WAL we use the logic above: 
```
        if (reader == null) {
          reader = newWalFileReader();
        } else {
          // reset read position after apply()
          reader.seekToFirstRecord();
        }
```
For the old WAL we always need the new instance. Additionally the try-with-resource block (223,224) is only applicable for the old file since it closes the reader after the block, for the recent WAL we need to keep it open.

src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
```diff
@@ -15,21 +15,27 @@
 
 package io.confluent.connect.hdfs.wal;
 
+import io.confluent.connect.hdfs.DataWriter;
```
kkonstantine, line 18: Most of our connectors use as import order as follows: 

```
import all other imports
<blank line>
import javax.*
import java.*
<blank line>
import io.confluent.*
<blank line>
import all other static imports
```
You might want to add this to you IDE to avoid a reshuffling as the one here.

src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
```diff
@@ -72,4 +78,235 @@ public void testTruncatedVersionWalFileRecovery() throws Exception {
...
+
+  private void addSampleEntriesToWAL(String topicsDir, WAL wal) throws IOException {
+    wal.append(WAL.beginMarker, "");
+    addSampleEntriesToWALNoMarkers(topicsDir, wal);
+    wal.append(WAL.endMarker, "");
+    wal.close();
+  }
+
+  private void addSampleEntriesToWALNoMarkers(String topicsDir, WAL wal) throws IOException {
+    for (int i = 0; i < 5; ++i) {
```
kkonstantine, line 301: Does it make sense to make the limit an argument of this method?

kkonstantine (commented):
Thanks for addressing the comments @dosvath 
Few more by me.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: Yes it's been working before, +1 to the same effect, I'll revert this.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,11 +188,156 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+    while (reader.next(key, value)) {
+      String keyName = key.getName();
+      if (keyName.equals(beginMarker)) {
+        tempFilenames.clear();
+        entryBlockStarted = true;
+      } else if (keyName.equals(endMarker)) {
+        if (entryBlockStarted && !tempFilenames.isEmpty()) {
+          // only save non-empty blocks
+          committedFilenames.clear();
+          committedFilenames.addAll(tempFilenames);
```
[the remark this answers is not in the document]
dosvath: I had synced with @erdody on this previously, the idea of `clear()` and `addAll()` here is to minimize the memory allocation. The effect is equivalent.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -183,11 +188,156 @@ private void commitEntriesToStorage(Map<WALEntry, WALEntry> entries) {
...
+   *
+   * @param reader the WAL file reader
+   * @return the last batch of entries, may be empty if the WAL
+   *         is empty or only contains empty BEGIN-END blocks
+   * @throws IOException error on reading the WAL file
+   */
+  private List<String> getLastFilledBlockFromWAL(Reader reader) throws IOException {
+    // In a WAL entry the temp filenames (keys) don't contain offset info,
+    // so we only need to track the committed filename (values).
+    List<String> committedFilenames = new ArrayList<>();
```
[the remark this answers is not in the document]
dosvath: It wouldn't be discarded if using the `clear()` and `addAll()` methods. If we decide to go with the `new ArrayList<>()` method above, then yes the `Collections.emptyList()` would be preferable.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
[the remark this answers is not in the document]
kkonstantine: The benefits are not as clear cut. For such optimizations it's worth checking the underlying implementations. 
* `addAll` creates a copy of the source array internally and assigns it to a local object. This local array will also be GC'ed once this method exits because it's copied to the destination array, it's not replacing it, because the method adds all the source elements to an existing array that potentially has more elements (not the case here). 
* `ArrayList(Collection ...)` creates such a copy but normally it assigns this to the destination array (the one that is constructed). So what is GC'ed is the old array in your code (as you've correctly observed). 

Therefore, besides a very vague and debatable advantage related to cache locality (which here might be even smaller because this is not a member variable array but still a local array within a method) we can't really be sure which approach works best. The number of arrays being GC'ed in the end is the same. And all that, without even saying that this is heavily dependent on whether the new arrays fit within the old one or the old one needs to get expanded more often than not. 

Given all that, unless this was surfaced as a bottleneck during our testing (which could include microbenchmarks with JMH if we wanted a quick assessment) or regular operation of the connector, I'd choose the most simple way. To me both work, so up to you. (As a last note, also worth saying that in modern jvm's allocation of new arrays has improved).

kkonstantine (approved):
Replied to the concern re: array allocation. 
In any case, probably either way works in the general case, so this PR LGTM

PR #559, opened 2021-04-30
MINOR: adopt java8 features and cleanup

levzem:

## Problem
the connector still uses java7 style when it can be updated to rely on java8 features
TPWTest was in the wrong package

## Solution
move TPWTest to the root package and and replace java7 language features with java8

- [ ] yes
- [x] no

## Test Strategy
run existing tests

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -37,7 +37,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
-import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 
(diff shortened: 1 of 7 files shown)


src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -874,46 +873,37 @@ private void setRetryTimeout(long timeoutMs) {
   }
 
   private void createHiveTable() {
```
dosvath, line 875: Do we have tests covering `createHiveTable()`, `alterHiveSchema()` and `addHivePartition()`, or verified the intended behavior of the updated lambda expressions other ways?

dosvath (commented):
Thanks @levzem, minor question.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
levzem: yes, we have multiple tests like these https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/test/java/io/confluent/connect/hdfs/avro/HiveIntegrationAvroTest.java
for each each dataformat

dosvath (approved):
LGTM

PR #561, opened 2021-05-04
CCMSG-1004: fix incorrect TopicPartitionWriter internal offset tracking

levzem:

## Problem
the connector [keeps track of offsets incorrectly](https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L883) by doing `offset = offset + recordCounter`

this makes the assumption that record offsets increment by 1 which is wrong
because offsets can increase by more than 1. imagine the scenario of a compacted topic which goes from 1 2 3 4 5 6 7 8 9 10 to 2 4 8 10, 10 records to 4, the starting offset is 1 + 10 = 11 is next offset which is correct but after compaction its 1 + 4 = 5, which is wrong. We would be repeating two extra records (8, 10).

## Solution
keep track of the latest offset by getting the largest end offset committed in the group of files committed

caveats: 
solution doesnt address if one file in a batch of commits fails which is relatively rare and only happens with a select few partitioners. 

lets say we use a partitioner that partitions 8 records into two different partitions (and by extension 2 different files) where records 3-6 go to file 1 and records 1-2,7-8 go to file 2. this can happen with a `FieldPartitioner` or the `TimeBasedPartitioner` when using the `RecordFieldTimestampExtractor` if the timestamps are out of order
from an EOS perspective, this group of files needs to be committed together as a transaction so that the next offset to commit will become 9. However they are not atomic transactions. file 1 can succeed to commit, but file 2 can fail to commit. rolling back is impossible because going back to offset 1 we will have duplicate records for 3-6, if we go to offset 7, we will be missing records 1-2.

 in the case if a file fails, we will just conservatively use the old solution of `offset = offset + recordCounter` since this cases also affects the current implementation which means the the risk is actually extremely low and on top of that the connector retries file commits infinitely so the correct offset will be written eventually

- [x] yes
- [ ] no

##### If yes, where?
potentially other sink storage connector i.e S3

##### Testing done:
- [x] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

 
backport to `5.0.x` because bug

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -828,7 +828,6 @@ public class TopicPartitionWriter {
     log.debug("Committing files");
     appended.clear();
 
-
     // commit all files and get the latest committed offset
     long latestCommitted = tempFiles.keySet().stream()
         .mapToLong(this::commitFile)
--- a/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
@@ -97,7 +97,7 @@ public class TopicPartitionWriter {
   private final Set<String> appended;
   private long offset;
   private final Map<String, Long> startOffsets;
-  private final Map<String, Long> offsets;
+  private final Map<String, Long> endOffsets;
   private final long timeoutMs;
   private long failureTime;
   private final StorageSchemaCompatibility compatibility;
@@ -201,7 +201,7 @@ public class TopicPartitionWriter {
     tempFiles = new HashMap<>();
     appended = new HashSet<>();
     startOffsets = new HashMap<>();
-    offsets = new HashMap<>();
+    endOffsets = new HashMap<>();
     state = State.RECOVERY_STARTED;
     failureTime = -1L;
     // The next offset to consume after the last commit (one more than last offset written to HDFS)
@@ -372,7 +372,7 @@ public class TopicPartitionWriter {
                         + "and end offsets {}",
                     tp,
                     startOffsets,
-                    offsets
+                    endOffsets
                 );
                 nextState();
                 // Fall through and try to rotate immediately
@@ -498,7 +498,7 @@ public class TopicPartitionWriter {
       exceptions.add(e);
     }
     startOffsets.clear();
-    offsets.clear();
+    endOffsets.clear();
 
     if (exceptions.size() != 0) {
       StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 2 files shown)


yanglei99 (approved):
lgtm

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -827,18 +827,25 @@ private void endAppend() {
   private void commitFile() {
     log.debug("Committing files");
     appended.clear();
+
+    long latestCommitted = -1;
     for (String encodedPartition : tempFiles.keySet()) {
-      commitFile(encodedPartition);
+      long endOffset = commitFile(encodedPartition);
+      latestCommitted = Math.max(latestCommitted, endOffset);
+    }
```
kkonstantine, line 835: How about?
`long lastCommitted = tempFiles.keySet().stream().mapToLong(this::commitFile).max().orElse(-1);`

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -720,7 +720,7 @@ private void writeRecord(SinkRecord record) {
     if (!startOffsets.containsKey(encodedPartition)) {
       startOffsets.put(encodedPartition, record.kafkaOffset());
     }
-    offsets.put(encodedPartition, record.kafkaOffset());
```
kkonstantine, line 723: if we were adding the correct end offset on the file, I wonder how this bug was demonstrating itself actually. 

Seems that the exported files would have the right start and end offsets even for compacted topics. I quickly looked at how we use the variable `offset` which would be the one to get misaligned/miscomputed and seems that this would only influence the offset we commit to the consumers topic. But we only depend on this offset as a fall back and we still recover from the max offset across the files on HDFS (we didn't use to depend at all but this has become a fallback since some time now). Have you noticed something different?

kkonstantine (commented):
Thanks for the fix @levzem 

We'll have to make the title a bit more specific as to what this fix is addressing exactly. Offset tracking was not incorrect in general. 

Related to that, I've added a question re: the use of `offset` and what this fix is supposed to address exactly.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -827,18 +827,25 @@ private void endAppend() {
   private void commitFile() {
     log.debug("Committing files");
     appended.clear();
-    for (String encodedPartition : tempFiles.keySet()) {
-      commitFile(encodedPartition);
+
```
kkonstantine, line 830:
```suggestion
```

kkonstantine (approved):
One last suggestion, but otherwise LGTM!

dosvath (approved):
LGTM

PR #567, opened 2021-05-13
CC-14520: fix jackson and hadoop CVEs

levzem:

## Problem
lots of CVEs

## Solution
resolve using various techniques
https://github.com/confluentinc/kafka-connect-storage-common/pull/185 should resolve a lot of them

jetty is being addressed by https://github.com/confluentinc/kafka-connect-storage-common/pull/190

only unresolvable CVEs are: 
- guava
- asl
- thrift

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,6 +110,12 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
+            <version>${jackson.databind.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>${jackson.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


kkonstantine (approved):
LGTM

pom.xml
```diff
@@ -54,7 +54,6 @@
 
     <properties>
         <confluent.maven.repo>http://packages.confluent.io/maven/</confluent.maven.repo>
-        <hadoop.version>2.7.3</hadoop.version>
```
dosvath, line 57: @levzem removing this will cause the common version to be used right? Is the version brought in from there identical? Trying to be careful it's not hadoop v 3.x.

levzem: thats the point of removing it. the version is `2.10.1` so we should be fine and I tested the changes

PR #568, opened 2021-05-19
CC-14408: Fix empty context assignment in start()

dosvath:
## Problem
The connect framework calls `open(Collection<TopicPartition> partitions)` in order to provide the assigned partitions for the task. However, the code is also attempting to fetch the assigned partitions information in `start()` via  `context.assignment()` from the `SinkTaskContext`. This only works for testing purposes as the context passed there is mocked. In the real calls when starting the connector, `context.assignment()` will come in empty in `start()` as the connect framework has not yet assigned the partitions. The partitions are only assigned on `open()`.

This is due to the fact that `context.assigment()` is not initialized until at least one `put()` call has returned, which means that referencing `context.assigment()` in `start()` is incorrect as it will always be empty. This has also resulted in some code blocks being skipped such as `syncwithHive()` where a Hive table is supposed to be created.

## Solution
Revise code to use partitions assigned in `open()` and remove the unused blocks from `start()`. Additionally, revise the tests to use `open()` for partition assignment. 

- [ ] yes
- [X] no

##### Testing done:
- [X] Unit tests
- [ ] Integration tests
- [ ] System tests
- [X] Manual tests

## Release Plan

 
Backporting to `10.0.x`.

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -15,6 +15,8 @@
 
 package io.confluent.connect.hdfs;
 
+import io.confluent.connect.storage.errors.HiveMetaStoreException;
+import java.util.stream.Collectors;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.Path;
@@ -36,7 +38,6 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -52,8 +53,6 @@ import java.util.concurrent.TimeUnit;
 import io.confluent.common.utils.SystemTime;
 import io.confluent.common.utils.Time;
 import io.confluent.connect.avro.AvroData;
-import io.confluent.connect.hdfs.filter.CommittedFileFilter;
-import io.confluent.connect.hdfs.filter.TopicCommittedFileFilter;
 import io.confluent.connect.hdfs.hive.HiveMetaStore;
 import io.confluent.connect.hdfs.hive.HiveUtil;
 import io.confluent.connect.hdfs.partitioner.Partitioner;
@@ -70,8 +69,6 @@ public class DataWriter {
 
   private final Map<TopicPartition, TopicPartitionWriter> topicPartitionWriters;
   private HdfsStorage storage;
-  private HashMap<String, String> logDirs;
-  private HashMap<String, String> topicDirs;
   private Format format;
   private RecordWriterProvider writerProvider;
   private io.confluent.connect.storage.format.RecordWriterProvider<HdfsSinkConnectorConfig>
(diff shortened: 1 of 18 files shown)


src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -380,50 +339,79 @@ public void recover(TopicPartition tp) {
...
       throw new ConnectException(e);
     }
   }
 
+
+  /**
+   * Get the latest file written for a topic. Go through the topic partition writers
+   * for a topic and make use of the already computed latest file in recover.
+   */
+  private Path getLatestFilePathForTopic(String topic) {
```
levzem, line 399: would be nice to move private methods below public ones

levzem: also add javadoc for the return and argument

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -380,50 +339,79 @@ public void recover(TopicPartition tp) {
...
+      FileStatus[] statuses = FileUtils.getDirectories(storage, new Path(topicDir));
+      for (FileStatus status : statuses) {
+        String location = status.getPath().toString();
+        if (!partitions.contains(location)) {
+          String partitionValue = getPartitionValue(location);
+          hiveMetaStore.addPartition(hiveDatabase, topic, partitionValue);
         }
       }
+    } catch (HiveMetaStoreException e) {
+      log.warn("Could not create Hive table: {}", e.getMessage());
```
levzem, line 388: why are we catching this here? shouldn't we fail?

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -380,50 +339,79 @@ public void recover(TopicPartition tp) {
...
+  }
+
+  /**
+   * Attempt to create a Hive table based on a file's schema from the topic.
+   *
+   * @param topic the name of the topic
+   * @param topicDir the directory of the topic
+   * @param latestSchema the schema of the latest file
+   */
+  private void attemptCreatingHiveTable(String topic, String topicDir, Schema latestSchema) {
```
levzem, line 375:
```suggestion
  private void createHiveTable(String topic, String topicDir, Schema latestSchema) {
```

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -144,18 +139,6 @@ public DataWriter(
...
-    for (TopicPartition tp : context.assignment()) {
-      String topicDir = connectorConfig.getTopicsDirFromTopic(tp.topic());
-      String logDir = connectorConfig.getLogsDirFromTopic(tp.topic());
-
-      topicDirs.put(tp.topic(), topicDir);
-      logDirs.put(tp.topic(), logDir);
-
-      createDir(topicDir);
-      createDir(topicDir + HdfsSinkConnectorConstants.TEMPFILE_DIRECTORY);
-      createDir(logDir);
```
levzem, line 156: where does this get created

src/main/java/io/confluent/connect/hdfs/DataWriter.java
dosvath: @levzem good point, we should fail to remain consistent with previous behavior.

levzem (approved):
LGTM, thanks @dosvath

PR #574, opened 2021-06-28
CCMSG-1133: Make hive table name configurable

erdody:
## Problem

Hive table name is currently the same as the topic name. We need it to be configurable.

## Solution

Add configuration property to customize hive table name, with support for `${topic}` substitution.

- [x] yes
- [ ] no

##### If yes, where?

 hdfs3

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests


erdody:
> It looks like @erdody hasn't signed our **C**ontributor **L**icense **A**greement, yet.
> 
> > The purpose of a CLA is to ensure that the guardian of a project's outputs has the necessary ownership or grants of rights over all contributions to allow them to distribute under the chosen licence.
> > [Wikipedia](http://en.wikipedia.org/wiki/Contributor_License_Agreement)
> 
> You can read and sign our full Contributor License Agreement [here](http://clabot.confluent.io/cla).
> 
> Once you've signed reply with `[clabot:check]` to prove it.
> 
> Appreciation of efforts,
> 
> clabot

[clabot:check]

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -141,7 +142,8 @@ public TopicPartitionWriter(
         null,
         null,
         null,
-        time
+        time,
+        "customHiveTableName-" + tp.topic()
```
kpatelatwork, line 146: why  "customHiveTableName-"?

src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
```diff
@@ -86,6 +86,7 @@ private Table constructAvroTable(
     Table table = newTable(database, tableName);
     table.setTableType(TableType.EXTERNAL_TABLE);
     table.getParameters().put("EXTERNAL", "TRUE");
+    // TODO not anymore, do we update the comment or do we need to update the path?
     // tableName is always the topic name
     String tablePath = hiveDirectoryName(url, config.getTopicsDirFromTopic(tableName), tableName);
```
kpatelatwork, line 91: we need to change the code because the tableName passed here is already having "prefix-${topic}" in it and if we read cofig.getTopicsDirFromTopic(tableName) then it's documentation says one more replacement so it would be wrong if we replace tableName there instead of topic.

Top level directory to store the data ingested from Kafka. Supports ``${topic}`` in the value, which will be "
      replaced by the actual topic name. Supports ``${0}``, ..., ``${n}`` in ................."

erdody:
Not sure why this is not compiling, it works locally. Will try again with the final release once https://github.com/confluentinc/kafka-connect-storage-common/pull/203 is approved.

pom.xml
```diff
@@ -114,7 +114,7 @@
         <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>kafka-connect-storage-hive</artifactId>
-            <version>${kafka.connect.storage.common.version}</version>
+            <version>10.2.0-20210704.081701-800</version>
```
kpatelatwork, line 117: any reasons to hardcode here instead of changing the value in the properties section?

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -91,6 +93,12 @@
   private static final String HADOOP_HOME_DOC = "The Hadoop home directory.";
   private static final String HADOOP_HOME_DISPLAY = "Hadoop home directory";
 
+  public static final String HIVE_TABLE_NAME_CONFIG = "hive.table.name";
+  public static final String HIVE_TABLE_NAME_DEFAULT = TOPIC_SUBSTITUTION;
+  private static final String HIVE_TABLE_NAME_DOC = "The hive table name to use. "
+          + "It supports the use of ${topic} to inject the corresponding topic name.";
```
kpatelatwork, line 99: should we tell the user that it must contain ${topic}

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -233,6 +241,18 @@ public static ConfigDef newConfigDef() {
           HADOOP_HOME_DISPLAY
       );
 
+      configDef.define(
```
kpatelatwork, line 244: I am not sure how this orderInGroup is used internally but as we are adding this property in the middle, will it cause any issues with the already materiazlied configs? Can you find out how orderInGroup is used and if its used then we should add this as the last config?

src/test/java/io/confluent/connect/hdfs/avro/HiveIntegrationAvroTest.java
```diff
@@ -244,21 +258,13 @@ public void testHiveIntegrationFieldPartitionerAvro() throws Exception {
 
     assertEquals(expectedColumnNames, actualColumnNames);
 
-    List<String> partitionFieldNames = connectorConfig.getList(
-        PartitionerConfig.PARTITION_FIELD_NAME_CONFIG
+    List<String> expectedPartitions = Arrays.asList(
+            partitionLocation(TOPIC, 16, "int"),
```
kpatelatwork, line 262: should we extract "int" as a variable called partitionFieldName as I had to read some code to figure this out:)?

src/test/java/io/confluent/connect/hdfs/avro/AvroHiveUtilTest.java
```diff
@@ -15,153 +15,17 @@
...
-
-    hdfsWriter.write(sinkRecords);
-    hdfsWriter.close();
-    hdfsWriter.stop();
-  }
-
-  private DataWriter createWriter(SinkTaskContext context, AvroData avroData){
-    return new DataWriter(connectorConfig, context, avroData);
+  @Override
+  protected HiveUtil createHiveUtil() {
```
kpatelatwork, line 28: Good Job on cleaning the duplicate code :)

src/test/java/io/confluent/connect/hdfs/avro/HiveIntegrationAvroTest.java
```diff
@@ -45,11 +47,34 @@
...
+
+  private final Map<String, String> localProps = new HashMap<>();
+  private final String hiveTableNameConfig;
+
+  public HiveIntegrationAvroTest(String hiveTableNameConfig) {
+    this.hiveTableNameConfig = hiveTableNameConfig;
+  }
+
+  @Parameters(name = "{index}: hiveTableNameConfig={0}")
+  public static Collection<Object[]> data() {
```
kpatelatwork, line 67: Very nice job here using this

kpatelatwork:
@erdody this is very good PR.  LGTM  from my side, but let's have @dosvath bless this because he has the most knowledge on HDFS.

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
erdody: I couldn't find anything that would break things if the order is updated, but just to be safe I moved it to the end of the group.

pom.xml
erdody: This is temporary until we can cut a release.
For some reason the -SNAPSHOT release is not working in Jenkins so I tried uploading this specific library manually.

src/test/java/io/confluent/connect/hdfs/avro/HiveIntegrationAvroTest.java
erdody: Any decent IDE would inline the name of the argument :)
image

src/main/java/io/confluent/connect/hdfs/HdfsSinkConnectorConfig.java
```diff
@@ -96,7 +96,7 @@
   public static final String HIVE_TABLE_NAME_CONFIG = "hive.table.name";
   public static final String HIVE_TABLE_NAME_DEFAULT = TOPIC_SUBSTITUTION;
   private static final String HIVE_TABLE_NAME_DOC = "The hive table name to use. "
-          + "It supports the use of ${topic} to inject the corresponding topic name.";
+          + "It requires the use of ${topic} to inject the corresponding topic name.";
```
kpatelatwork, line 99: instead of requires should we say "It must contain ${topic} ...."

dosvath (commented):
Thanks @erdody! one question

src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java
```diff
@@ -80,14 +84,15 @@ private Table constructAvroTable(
...
+      String topic
   )
       throws HiveMetaStoreException {
     Table table = newTable(database, tableName);
     table.setTableType(TableType.EXTERNAL_TABLE);
     table.getParameters().put("EXTERNAL", "TRUE");
-    // tableName is always the topic name
-    String tablePath = hiveDirectoryName(url, config.getTopicsDirFromTopic(tableName), tableName);
+
+    String tablePath = hiveDirectoryName(url, config.getTopicsDirFromTopic(topic), topic);
```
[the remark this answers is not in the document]
erdody: I don't think there will be issues with hive, we have tests covering it. The reason to keep them separate is that location is used in other places (e.g. [TopicPartitionWriter](https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java#L341)) and besides the extra effort it might be confusing to write to different places depending on whether hive integration is turned on or not.

dosvath (approved):
LGTM, thanks @erdody! Let's get the storage-common released and the version here updated.

theworkshopcom:
hi @erdody , @dosvath , @kpatelatwork , we would like to use this feature to configure a hive table name to be `short_name` for  a Kafka topic with name: `a_very_long_name_that_does_not_make_sense_on_hive`. 

Are we saying that we cannot do this, and we **must** always include ${topic} subsctitution on the name? 

Why is that restriction in place, it doesn't make sense if I understand this correctly

PR #579, opened 2021-07-27
CCMSG-1014: Do not leak fs in io.confluent.connect.hdfs.wal.WALFile.Writer#Writer().

avocader:
## Problem
HDFS client has a cache - `org.apache.hadoop.fs.FileSystem#CACHE`.
It's not automatically cleaned, it requires an explicit cleanup.
In the HDFS sink connector, some actions create new entries in the cache and some clean entries from it(added comments in the code, to make review easier, will remove before merging).

`io.confluent.connect.hdfs.wal.FSWAL#FSWAL` does actions, that create cache entries:
- `append()`
- `acquireLease()`
- `apply()`
- `truncate()`

Whenever we call `org.apache.hadoop.fs.FileSystem#newInstance(java.net.URI, org.apache.hadoop.conf.Configuration)`, it creates one entry in `org.apache.hadoop.fs.FileSystem#CACHE` so it has to be closed when not needed anymore.

There is a `catch` clause in `io.confluent.connect.hdfs.wal.WALFile.Writer#Writer`, that closes `fs` in case there is `RemoteException`.
But, in some cases, it can be another exception(e.g. `java.io.EOFException`, when HDFS is down), we are not catching it and re-throwing an exception.

## Solution
Catch `Exception` instead of `RemoteException` so that we never leak `fs`.

- [ ] yes
- [ ] no

##### Testing done:
- [x] Unit tests
- [x] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
+++ b/src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
@@ -103,7 +103,8 @@ public class FSWALTest extends TestWithMiniDFSCluster {
     Object cache = cacheField.get(Object.class);
     Field cacheMapField = cache.getClass().getDeclaredField("map");
     cacheMapField.setAccessible(true);
-    Map cacheMap = (Map) cacheMapField.get(cache);
+    //suppressing the warning since org.apache.hadoop.fs.FileSystem.Cache.Key has package-level visibility
+    @SuppressWarnings("rawtypes") Map cacheMap = (Map) cacheMapField.get(cache);
     cacheField.setAccessible(false);
     cacheMapField.setAccessible(false);
     return cacheMap.size();
--- a/src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
+++ b/src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
@@ -15,14 +15,11 @@
 
 package io.confluent.connect.hdfs.wal;
 
-import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.kafka.common.TopicPartition;
 import org.junit.Test;
 
 import java.io.OutputStream;
-import java.lang.reflect.Field;
-import java.util.Map;
 
 import io.confluent.connect.hdfs.FileUtils;
 import io.confluent.connect.hdfs.TestWithMiniDFSCluster;
@@ -95,18 +92,5 @@ public class FSWALTest extends TestWithMiniDFSCluster {
     // make sure org.apache.hadoop.fs.FileSystem.CACHE
     // doesn't grow when acquireLease re-throws an exception
     assertEquals(fileSystemCacheSizeBefore, getFileSystemCacheSize());
-  }
-
-  private int getFileSystemCacheSize() throws Exception {
-    Field cacheField = FileSystem.class.getDeclaredField("CACHE");
-    cacheField.setAccessible(true);
-    Object cache = cacheField.get(Object.class);
-    Field cacheMapField = cache.getClass().getDeclaredField("map");
-    cacheMapField.setAccessible(true);
-    //suppressing the warning since org.apache.hadoop.fs.FileSystem.Cache.Key has package-level visibility
-    @SuppressWarnings("rawtypes") Map cacheMap = (Map) cacheMapField.get(cache);
-    cacheField.setAccessible(false);
-    cacheMapField.setAccessible(false);
-    return cacheMap.size();
   }
 }
--- a/src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
+++ b/src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
@@ -15,17 +15,20 @@
 
 package io.confluent.connect.hdfs.wal;
 
-import io.confluent.connect.hdfs.FileUtils;
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.kafka.common.TopicPartition;
 import org.junit.Test;
 
+import java.io.OutputStream;
+import java.lang.reflect.Field;
+import java.util.Map;
+
+import io.confluent.connect.hdfs.FileUtils;
 import io.confluent.connect.hdfs.TestWithMiniDFSCluster;
 import io.confluent.connect.hdfs.storage.HdfsStorage;
-import io.confluent.connect.hdfs.storage.Storage;
 
-import java.io.OutputStream;
-
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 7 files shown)


src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -99,6 +99,7 @@ public void acquireLease() throws ConnectException {
           throw new ConnectException(e);
         }
       } catch (IOException e) {
+        close();
```
kpatelatwork, line 102: do we have to call close in the RemoteExceptionBlock also specially when we are throwing ConnectException?

kpatelatwork: Also what about below block where you acquire the writer but we are throwing    

```
 if (sleepIntervalMs >= WALConstants.MAX_SLEEP_INTERVAL_MS) {
      throw new ConnectException("Cannot acquire lease after timeout, will retry.");
}
```

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
```diff
@@ -816,7 +816,7 @@ private void appendToWAL() {
 
   private void beginAppend() {
     if (!appended.contains(WAL.beginMarker)) {
-      wal.append(WAL.beginMarker, "");
+      wal.append(WAL.beginMarker, ""); //this creates a cache entry
```
dosvath, line 819: `wal.append` in the `FSWAL.java` class calls two methods on the writer
```
      writer.append(key, value);
      writer.hsync();
```
Does this create two cache entries? above you mention a single one. If two entries are created and only one is cleaned that could also mean accumulation over time.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
dosvath: +1 that `close()` may need to be called in the `RemoteExceptionBlock` before throwing a `ConnectException`.

dosvath: One additional thought about calling `close()` here (that may not be an issue) is that `close()` also closes the reader if it's non-null, but in this `acquireLease()` method we only call methods on the writer.

dosvath (approved):
Thanks @avocader I think we're close on resolving this.

dosvath (changes requested):
Approved by accident, pending comments above.

src/main/java/io/confluent/connect/hdfs/TopicPartitionWriter.java
avocader: @dosvath calling `writer.append(key, value);` and `writer.hsync();` doesn't create 2 cache entries(the size of `org.apache.hadoop.fs.FileSystem.Cache#map` doesn't change when I step over these lines in debugger).
Calling `wal.append(WAL.beginMarker, "");` creates one cache entry, this is a stack trace that I was able to get while running `HdfsSinkTaskTest`:
```
getInternal:3310, FileSystem$Cache (org.apache.hadoop.fs)
getUnique:3270, FileSystem$Cache (org.apache.hadoop.fs)
newInstance:528, FileSystem (org.apache.hadoop.fs)
<init>:146, WALFile$Writer (io.confluent.connect.hdfs.wal)
createWriter:78, WALFile (io.confluent.connect.hdfs.wal)
acquireLease:74, FSWAL (io.confluent.connect.hdfs.wal) --> we create a writer here
append:57, FSWAL (io.confluent.connect.hdfs.wal)
createWALs:365, HdfsSinkTaskTest (io.confluent.connect.hdfs)
testSinkTaskStartWithRecovery:233, HdfsSinkTaskTest (io.confluent.connect.hdfs)
...
```
So, essentially, creating one writer creates a one cache entry, and leaving unclosed writers(e.g. we catch and re-throw and exception in `io.confluent.connect.hdfs.wal.FSWAL#acquireLease`) blows up the cache.

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
avocader: Good catch @kpatelatwork! You are right, whenever we call
```
writer = WALFile.createWriter(conf, Writer.file(new Path(logFile)),
                                        Writer.appendIfExists(true));
```
it creates a cache entry, and we should always close it if applicable in case we are throwing an exception.
I moved `close()` to a `finally` block.

@dosvath right, `acquireLease()` only creates `writer`. Do you think it makes sense to split `close()` into `closeReader()` and `closeWriter()` and call `closeWriter()` from `acquireLease()`?

src/main/java/io/confluent/connect/hdfs/wal/FSWAL.java
```diff
@@ -109,6 +108,8 @@ public void acquireLease() throws ConnectException {
             ),
             e
         );
+      } finally {
```
kpatelatwork: Hey @avocader  moving to finally will close the reader/writer even when the lease is successfully acquired.  I do not see that in the old code.  Are we sure this is the right fix?

avocader: I deleted `close()` call, so not relevant anymore.

src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
```diff
@@ -72,4 +75,38 @@ public void testTruncatedVersionWalFileRecovery() throws Exception {
     FSWAL wal = new FSWAL("/logs", tp, storage);
     wal.acquireLease();
   }
+
+  @Test
+  public void testAcquireLeaseThrowsException() throws Exception {
```
kpatelatwork, line 80: Based on your new fix, Do we need a new test that is testing the Writer instead of acquireLease.  We should leave this test, I mean this is a good test.

src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
```diff
@@ -72,4 +75,38 @@ public void testTruncatedVersionWalFileRecovery() throws Exception {
...
+    setUp();
+    HdfsStorage storage = new HdfsStorage(connectorConfig, url);
+    TopicPartition tp = new TopicPartition("mytopic", 123);
+    //stop HDFS cluster so that WALFile.createWriter() throws an exception
+    cluster.shutdown();
+    FSWAL wal = new FSWAL("/logs", tp, storage);
+    int fileSystemCacheSizeBefore = getFileSystemCacheSize();
+    for (int i = 0; i < 10; i++) {
+      try {
+        wal.acquireLease();
```
kpatelatwork, line 90: One other question is that you did found out the real issue with Wal file and suggested customer to fix the permissions, don't we need that fix also?

src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
avocader: Added a similar test - `io.confluent.connect.hdfs.wal.WALFileTest#testHdfsIsDown`.

src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
avocader: It's a generic issue, that could be caused by multiple reasons - lack of HDFS privileges, HDFS is down, etc.
The PR fixes the generic issue of leaking `fs` instance in case there's an exception, different from `RemoteException`.
I also logged a separate ticket to improve permissions validation - https://confluentinc.atlassian.net/browse/CCMSG-1297

dosvath (approved):
LGTM, thanks @avocader, let's get +1 from @kpatelatwork as well.

src/test/java/io/confluent/connect/hdfs/wal/FSWALTest.java
kpatelatwork: Thanks it look good

kpatelatwork (approved):
@avocader This PR looks good. Thanks a lot for your persistence on the issue.

LGTM

PR #610, opened 2022-02-25
CC-16542 Add log4j2 core

cjolivier01:
- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


kpatelatwork (approved):
LGTM

Just curious if there was any reason we changed the variable name?

cjolivier01:
> LGTM
> 
> Just curious if there was any reason we changed the variable name?

in order to differentiate it with log4j versions which get replaced by confluent-log4j, which is API 1.2.x

Issue #331, opened 2018-05-29
Over Hive `avro.schema.literal` max size due to so long Avro schema 


daigorowhite:
Hi team, I have one issue with long Avro schema with `kafka-connect-hdfs` hive integration.

When I try to sink long schema table into HDFS with `kafka-connect-hdfs`.

Success to put data into HDFS , but I got this error when I throw query
```
0: jdbc:hive2://localhost:10000/default> select * from db.table limit 1 ;
Error: Error while compiling statement: FAILED: RuntimeException MetaException(message:org.apache.hadoop.hive.serde2.SerDeException Encountered exception determining schema. Returning signal schema to indicate problem: org.codehaus.jackson.JsonParseException: Unexpected end-of-input: was expecting closing quote for a string value
 at [Source: java.io.StringReader@2c87f90b; line: 1, column: 6001]) (state=42000,code=40000)
```

I investigated root cause of this, and it is caused the Hive meta data param `varchar` size.
https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java#L69
https://github.com/confluentinc/kafka-connect-hdfs/blob/master/src/main/java/io/confluent/connect/hdfs/avro/AvroHiveUtil.java#L95

```
mysql> describe TABLE_PARAMS;
+-------------+---------------+------+-----+---------+-------+
| Field       | Type          | Null | Key | Default | Extra |
+-------------+---------------+------+-----+---------+-------+
| TBL_ID      | bigint(20)    | NO   | PRI | NULL    |       |
| PARAM_KEY   | varchar(256)  | NO   | PRI | NULL    |       |
| PARAM_VALUE | varchar(4000) | YES  |     | NULL    |       |
+-------------+---------------+------+-----+---------+-------+
```

Do you have any solution/idea for this in `kafka-connect-hdfs`?

rhauch:
I've seen this workaround: log into Hive Metastore DB, then run `alter table SERDE_PARAMS MODIFY PARAM_VALUE VARCHAR(40000);`

Probably want to pick a realistic size, though.

daigorowhite:
It is one way to avoid this problem, but I think it is depended Hive Setting side and could be shorten again by Hive upgrade.
How about just use `avro.schema.url` to have un-limited size `.avro` ?

rhauch:
[...]

Issue #311, opened 2018-04-03
Failed to build from source


lnquy:
Hi all,
I tried to build from source code of 4.0.0 release but there're two problems in `pom.xml` file which still remained in HEAD.  
I noticed these two problems already reported in [schema-registry](https://github.com/confluentinc/schema-registry/issues/630) but not on this repo so I just put it here.  

1. Maven failed to resolve the `confluent.maven.repo` property for Confluent repository's URL.
```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-resolvable parent POM for io.confluent:rest-utils-parent:[unknown-version]: Could not transfer artifact io.confluent:common:pom:4.0.0 from/to confluent (${confluent.maven.repo}): Cannot access ${confluent.maven.repo} with type default using the available connector factories: BasicRepositoryConnectorFactory and 'parent.relativePath' points at wrong local POM @ line 8, column 13
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project io.confluent:rest-utils-parent:[unknown-version] (/home/lnquy/workspace/projects/java/rest-utils/pom.xml) has 1 error
[ERROR]     Non-resolvable parent POM for io.confluent:rest-utils-parent:[unknown-version]: Could not transfer artifact io.confluent:common:pom:4.0.0 from/to confluent (${confluent.maven.repo}): Cannot access ${confluent.maven.repo} with type default using the available connector factories: BasicRepositoryConnectorFactory and 'parent.relativePath' points at wrong local POM @ line 8, column 13: Cannot access ${confluent.maven.repo} using the registered transporter factories: WagonTransporterFactory: Unsupported transport protocol -> [Help 2]
```

2. Missing Confluent repository for plugins cause failed to find io.confluent:build-tools:jar.
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check (validate) on project kafka-connect-hdfs: Execution validate of goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:check failed: Plugin org.apache.maven.plugins:maven-checkstyle-plugin:2.17 or one of its dependencies could not be resolved: Failure to find io.confluent:build-tools:jar:4.0.0 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced -> [Help 1]
```

[...]

PR #613, opened 2022-04-05
Fix pom.xml repository url substitution property failure.

geniusjoe:
## Problem
I'm new to this project. 
When I forked the project to the local computer,my IDEA ide throw an exception about cannot resolve parent project.

shortcut1

```
Could not transfer artifact io.confluent:kafka-connect-storage-common-parent:pom:10.1.2 from/to confluent (${confluent.maven.repo}): Cannot access ${confluent.maven.repo} with type default using the available connector factories: BasicRepositoryConnectorFactory
```
My IDEA cannot resolve the parent dependency `io.confluent:kafka-connect-storage-common-parent` although it can be found on [confluent site](http://packages.confluent.io/maven/io/confluent/kafka-connect-storage-common-parent/). Therefore I think it's not caused by SnapShot version.

image

## Steps to repreduce
Just simply delete package `.m2\repository\io\confluent\kafka-connect-storage-common-parent\10.1.2` from local .m2 dependency cache, git checkout to 10.0.0.6 branch and then reload project.

## Solution

An issue about this exception has been mentioned in [#331](https://github.com/confluentinc/kafka-connect-hdfs/issues/311).
According to these StackOverflow [thread1](https://stackoverflow.com/questions/36111219/maven-property-not-resolved-in-parent-tag/36114683#36114683) [thread2](https://stackoverflow.com/questions/59650854/maven-property-substitution-for-repository-url-not-happening-if-parent-pom-is-de) :

> I assume the error happens when parent pom is being downloaded from the repo? That would make sense because to build the whole list of properties you'd need to download the parent and build an Effective POM. So you can't apply properties because parent hasn't been downloaded and you can't download parent because properties haven't been applied. That's my take on why it doesn't work, but I didn't run any experiments to confirm the hypothesis.

Maven will substitude property after parent pom has been parsed, so if parent pom is from remote repo and repo url is not written explicit, maven cannot download it.

- [x] yes
- [ ] no

##### If yes, where?
Any other version using substitution in pom.xml repository url.

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [x] Manual tests


geniusjoe:
@cjolivier01 @kpatelatwork @kkonstantine 
Any advice you could give would be much appreciated.

orinciog:
Hi! Your fix also worked for me. Thank you!

PR #614, opened 2022-04-27
FIX: Set fs.automatic.close to false in Hadoop configuration

snehashisp:
## Problem
The Hadoop DFS client auto closes and created clients, when the JVM shuts down. This can happen prior to the connector being closed, which requires these clients to be in open state for deletion of temporary files on close. Note that this error only comes into the picture when the connect worker is stopped/restarted. 

## Solution
Hadoop has a configuration `fs.automatic.close` which is by default true. Setting this as `false` will disable this behaviour. This is safe to do as we are anyways not relying on the auto close feature for closing all the clients. The connector lifecycle has appropriate routines for closing these clients on exit since it needs to rely on the lifecycle during a a connector delete, where the JVM shut down hooks are not executed. The connector should always operate with `fs.automatic.close` as false. 

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -132,6 +132,17 @@ public class DataWriter {
       hadoopConfiguration.addResource(new Path(config.hadoopConfDir() + "/hdfs-site.xml"));
     }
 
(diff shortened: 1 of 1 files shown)


kpatelatwork (approved):
LGTM

PR #628, opened 2022-09-02
Added new property KERBEROS_REFRESH_TICKET for HdfsSinkConnectorConfi…

joshbigdaddy:
…g that allows the possibility of never refreshing the token

## Problem
Needed to never refresh kerberos token due to kerberos stopping the pipeline each week
## Solution
Created a common solution that allows users not to use refresh of kerberos tickets at all

- [ ] yes
- [x] no

## Test Strategy
I have tested that the configuration applies correctly in the configuration test

##### Testing done:
- [x ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests

## Release Plan

Only releases a new config feature, so it should be harmless

Mergin to master
 
I have put this config value as true by default so it is safe

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -235,8 +235,7 @@ public class DataWriter {
 
     initializeTopicPartitionWriters(context.assignment());
   }
-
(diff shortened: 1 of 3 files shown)


joshbigdaddy:
Hi, I have already solved all the problems and checks have been passed

PR #630, opened 2022-09-26
Remove ant dependency

snehashisp:
## Problem
CVE https://confluentinc.atlassian.net/browse/CCMSG-1901

## Solution
Remove ant dependency. It may be good to remove the ant dependency from parent but raising this PR so that tests can run and we can check if we can actually remove the dependency. 

- [ ] yes
- [ ] no

##### Testing done:
- [ ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ ] Manual tests


arihant-confluent (approved):
LGTM

Issue #325, opened 2018-05-10
Can't have multiple Hadoop clients in same worker


thomaskwscott:
It's not currently using multiple Connectors to different Hadoop environments within a single Connect worker. We suspect the problem is here:

https://github.com/confluentinc/kafka-connect-hdfs/blob/94b539dbb70aaf18d4436d2413b784e25b55f205/src/main/java/io/confluent/connect/hdfs/DataWriter.java#L168.

Where UserGroupInformation is shared between all instances of the Hadoop client. This could be addressed using ClassLoader isolation as per:

https://stackoverflow.com/questions/36845701/why-setconfigurationconfiguration-conf-method-of-usergroupinformation-class-is

rayokota:
There is a wait to solve the problem without classloaders.  You would use the `UserGroupInformation.loginUserFromKeytabAndReturnUGI()` call:  https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/security/UserGroupInformation.html#loginUserFromKeytabAndReturnUGI(java.lang.String,%20java.lang.String)   Also see “Multiple UGIs” here: https://bryanbende.com/development/2016/03/20/learning-kerberos-and-ugi

nameof:
its not，what thomaskwscott  mean is, "using multiple Connectors to different Hadoop environments". ```UserGroupInformation.loginUserFromKeytabAndReturnUGI()``` is used to solve multiple user connects to same hadoop environment(the same configuration via ```UserGroupInformation.setConfiguration```) in single  jvm. <br/>
According to StackOverflow，if we want to access multiple cluster from single JVM then we can use multiple classloaders to get around the problem of setting multiple configuration from single JVM.

PR #674, opened 2023-11-25
Multiple keytab kerberos issue

zeun0725:
## Problem
If multiple key tabs are distributed on the connect server, authentication errors may occur.

```
      UserGroupInformation.loginUserFromKeytab(principal, connectorConfig.connectHdfsKeytab()); // problem point
      final UserGroupInformation ugi = UserGroupInformation.getLoginUser();  // problem point
      log.info("Login as: " + ugi.getUserName());

      isRunning = true;
      ticketRenewThread = new Thread(() -> renewKerberosTicket(ugi));
```
example) A connector uses the test user keytab (write path: /hdfs/user/test), and B connector uses the test2 user keytab (write path: /hdfs/user/test2).
If both connectors are restarted due to certain circumstances, different key tab information can be imported with thread-based behavior.
If the B connector calls loginUserFromKeytab('[test2@EXAMPLE.COM](mailto:test2@EXAMPLE.COM)' , '/../test2.keytab') when the A connector calls getLoginUser(), A login as test2 user. So you can face the error as below.
Caused by: org.apache.ranger.authorization.hadoop.exceptions.RangerAccessControlException: Permission denied: user=test2, access=EXECUTE, inode="/hdfs/user/test"

## Solution
```      
final UserGroupInformation ugi = UserGroupInformation.loginUserFromKeytabAndReturnUGI(principal, connectorConfig.connectHdfsKeytab());
```
When receiving an ugi instance, it was changed to authenticate with a keytab and receive it.
Change to use loginUserFromKeytabAndReturnUGI() func.

- [ ] yes
- [ X ] no

##### Testing done:
- [ X ] Unit tests
- [ ] Integration tests
- [ ] System tests
- [ X ] Manual tests

### DIFF
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -274,7 +274,7 @@ public class DataWriter {
           hostname
       );
       UserGroupInformation.loginUserFromKeytab(principal, connectorConfig.connectHdfsKeytab());
-      final UserGroupInformation ugi = UserGroupInformation.getLoginUser();
+      final UserGroupInformation ugi = UserGroupInformation.loginUserFromKeytabAndReturnUGI(principal, connectorConfig.connectHdfsKeytab())
       log.info("Login as: " + ugi.getUserName());
 
       isRunning = true;
--- a/src/main/java/io/confluent/connect/hdfs/DataWriter.java
+++ b/src/main/java/io/confluent/connect/hdfs/DataWriter.java
@@ -274,7 +274,7 @@ public class DataWriter {
           hostname
       );
       UserGroupInformation.loginUserFromKeytab(principal, connectorConfig.connectHdfsKeytab());
-      final UserGroupInformation ugi = UserGroupInformation.loginUserFromKeytabAndReturnUGI(principal, connectorConfig.connectHdfsKeytab())
+      final UserGroupInformation ugi = UserGroupInformation.loginUserFromKeytabAndReturnUGI(principal, connectorConfig.connectHdfsKeytab());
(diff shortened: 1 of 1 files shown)


lala7573:
@kkonstantine  @snehashisp 
Could you please review this PR? Related to this issue https://github.com/confluentinc/kafka-connect-hdfs/issues/325

@zeun0725 
Please sign the cla and make this valid pr.

src/main/java/io/confluent/connect/hdfs/DataWriter.java
```diff
@@ -274,7 +274,7 @@ private void configureKerberosAuthentication(Configuration hadoopConfiguration)
           hostname
       );
       UserGroupInformation.loginUserFromKeytab(principal, connectorConfig.connectHdfsKeytab());
-      final UserGroupInformation ugi = UserGroupInformation.getLoginUser();
+      final UserGroupInformation ugi = UserGroupInformation.loginUserFromKeytabAndReturnUGI(principal, connectorConfig.connectHdfsKeytab());
```
lala7573, line 277: If the line exceeds 100 characters, mvn checkstyle error occurs, so it would be a good idea to change it like this.

```suggestion
      final UserGroupInformation ugi = UserGroupInformation.loginUserFromKeytabAndReturnUGI(
            principal, 
            connectorConfig.connectHdfsKeytab()
      );
```

lala7573 (changes requested):

lala7573:
@zeun0725 
I build this and tested it.
It seems that there are other problems besides this issue. This changes only affect on relogin.
When I try to create a connector with different user, lease permission error still occurs.

zeun0725:
@lala7573 
Can you share the test code or config value?

